PR #2, opened 2015-06-04
Include option to prefix file reports

mrfyda:


rtfpessoa:
Looks good to me. If compiles you can merge.

PR #13, opened 2015-11-22
Fix printed version

### LINKED ISSUE #8: codacy-coverage-reporter prints wrong version
When running with --help flag, it prints 

```
codacy-coverage-reporter 1.0.0
Usage: codacy-coverage-reporter [options]

  -l <value> | --language <value>
        foo is an integer property
  -t <value> | --projectToken <value>
        your project API token
  -r <value> | --coverageReport <value>
        your project API token
  --codacyApiBaseUrl <value>
        the base URL for the Codacy API
  --prefix <value>
        the project path prefix
  --help
        prints this usage text
```

When the actual version is 1.0.4
This version should be fetched from the pom or something. This way you have to manually change in every release

https://github.com/codacy/codacy-coverage-reporter/blob/master/src/main/scala/com/codacy/CodacyCoverageReporter.scala#L65

pedrorijo91:
Instead of harcoded string (previously "1.0.0" since forever), now getting the from `build.sbt`

From: http://stackoverflow.com/questions/8957025/sbt-including-the-version-number-in-a-program

Closes #8

### DIFF
--- a/src/main/scala/com/codacy/CodacyCoverageReporter.scala
+++ b/src/main/scala/com/codacy/CodacyCoverageReporter.scala
@@ -62,7 +62,7 @@ object CodacyCoverageReporter {
 
   def buildParser = {
     new scopt.OptionParser[Config]("codacy-coverage-reporter") {
-      head("codacy-coverage-reporter", "1.0.0")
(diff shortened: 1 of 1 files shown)


mrfyda:
LGTM

PR #14, opened 2015-12-02
updated troubleshooting steps

amitmawkin:
based on discussion with Codacy support ,Thanks to Rafael from Codacy for helping

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,3 +60,14 @@ codacy-coverage-reporter -l Java -r PATH_TO_COVERAGE/coverage.xml
 
 2015-11-20 04:07:00,639 [error] com.codacy Failed to upload report: Not Found
 ```
+Even after doing all of the above troubleshooting steps in case you still encounter the same error 
+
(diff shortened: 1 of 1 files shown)


pedrorijo91:
:+1: thanks !

Nevertheless, I would rather not to use the \<code> tags and use markdown typical syntax with \``` 
And I think this information would make more sense in a section 'Usage', indicating the full path should be specified, as long with information about available flags.

Will wait for @machadoit feedback

machadoit:
Yeah looks great, thanks! Like @pedrorijo91 said, can you use \``` instead of the \<code\> tag?
The 'Usage' is our 'Updating Codacy', we will review this documentation if it's not clear, but for now let the info in the Troubleshooting as you did. :+1:

amitmawkin:
Done the changes, please review

machadoit:
Still missing the last line ;)

amitmawkin:
something weird let me submit another one

PR #37, opened 2016-11-03
Bump version for development iteration

pedrorijo91:


build.sbt
```diff
@@ -2,7 +2,7 @@ import Dependencies._
 
 name := "codacy-coverage-reporter"
 
-version := "1.0.5"
+version := "1.1.0-SNAPSHOT"
```
rtfpessoa: We do not use this version for anything. The build server injects the version every build and it is incremental.

pedrorijo91: what's the purpose then? should we delete it?

rtfpessoa: No, just leave it like it is.

Or find a proper plugin that not only increments it but also commits the change :smile:

PR #39, opened 2016-11-04
add alternative gradle task

MrRamych:


pedrorijo91:
Thanks for the contribution @MrRamych !

Can you just clarify us in some stuff?

- The existing gradle task didn't work for you? Can you tell us why?
- The task you added seems very similar to the one already existing on the README, but we never worked with gradle. Can you explain us the differences?

MrRamych:
It is shorter, no useless code, it works.
Existing task is much bigger and is not full - some code is missing

rtfpessoa:
@MrRamych sorry for the delay, this tocket got lost. Can you extend in "some code is missing"?

MrRamych:
```gradle
configurations {
     providedRuntime
     codacy
 }
dependencies {
    codacy group: 'com.codacy', name: 'codacy-coverage-reporter', version: '1.0.7'
}
```
Also user need to install reporter somehow. I could not figure out how to do it. But my solution works without any additional problems.

rtfpessoa:
Ok, thanks for the contribution.

PR #51, opened 2017-04-18
Adding a reference to the Codacy Gradle plugin.

ddimtirov:
I have added a reference to the new plugin. As this plugin subsumes the work of the previous two Gradle solutions by @MrRamych and @tompahoward, you may want to rewrite the text to reduce/remove the details and just provide credit. As it is now, having 3 different solutions without clear reccommendation can be confusing to a casual reader.

I haven't done it as it is a project owner's decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,20 @@ task sendCoverageToCodacy(type: JavaExec, dependsOn: jacocoTestReport) {
     ]
 }
 ```
+___
+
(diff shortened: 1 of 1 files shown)


rtfpessoa:
@ddimtirov please move your plugin to the top of the gradle tasks. I think keeping the tasks is a good idea because someone might be interested in fully controlling the code. And since they already exists I do not see any advantage in removing them.

ddimtirov:
Nevermind.

PR #57, opened 2017-07-14
CodacyCoverageReporter: Add support for AWS CodeBuild

thaddeusdiamond-smartercloud:
AWS CodeBuild doesn't copy out the .git directory (known issue discussed
here: https://forums.aws.amazon.com/thread.jspa?threadID=244197).
However, they do provide us with an environment variable we
can use.  Put this detection in the CodacyCoverageReporter so that we
can auto-resolve this and not have to use the --commitUUID flag.

[ Documentation: Updated the README ]
[ Testing: Build ]


pedrorijo91:
Close #58

PR #77, opened 2018-03-18
Update README.md

aharonha:
Getting the latest version from public repository (rather then GitHub releases)

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,31 @@ Just follow the [instructions on his repository](https://github.com/halkeye/coda
 If you want to use codacy with Travis CI and report coverage generated from your tests run in Travis, update your .travis.yml to include the following blocks:
 
 ```yaml
+sudo: required
 before_install:
-  - sudo apt-get install jq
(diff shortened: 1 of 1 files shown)


rtfpessoa:
@aharonha thanks for your contribution.

Overall looks good but I would prefer if we did not remove the previous option.
Can you add an extra option after the other with your suggestion?

rtfpessoa:
@aharonha did you see my comment?

rtfpessoa:
Closing this for now due to inactivity. In the meanwhile if you see my comment we can re-open this.

aharonha:
Actually the first time I made the pull request , the script worked.
After the latest release, my build was starting to fail.
The root cause is the change in deployment you made, making the latest
version available on Maven central repository. On the other hand, in
GitHub, which the original script was depending on, started to fail. I
believe it has to do with the fact that no jars was uploaded to GitHub
since 3.0

So, if the latest version is both available on central and it's fixing an
issue that the latest version available on GitHub meets, the latest version
should be downloaded from Central only.

(making the old script useless)


בתאריך יום ב׳, 26 במרץ 2018, 17:34, מאת Rodrigo Fernandes ‏<
notifications@github.com>:

PR #81, opened 2018-04-30
Allow the report format to be specified

DReigada:


DReigada:
We decided to leave the selection of the parser automatic

PR #84, opened 2018-05-29
Updated coverage parser

DReigada:


src/test/scala/com/codacy/rules/ConfigurationRulesTest.scala
```diff
@@ -34,7 +34,7 @@ class ConfigurationRulesTest extends FlatSpec with Matchers {
       case config: ReportConfig =>
         val result = components.reportRules.coverageWithTokenAndCommit(config)
 
-        result should be(Left("no parser for Scala"))
+        result should be(Left("could not parse report, unrecognized report format"))
```
rtfpessoa: Wasn't this suppose to say the report format names so people would know what we tried?

PR #109, opened 2018-11-22
Support coverage over multi modules

nenick:
First tests has passed, the merged coverage report for multiple modules was uploaded and analysed successfully.

use it as
`codacy-coverage-reporter report -r coverage.xml --multi-module app/src/main/java,lib/src/main/scala`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ It might also be possible to merge the reports before uploading them to Codacy, 
 For coverage reports including multiple modules we need the correct path for matching the covered files.
 
 ***Example***
-1. `codacy-coverage-reporter report -l Java -r report1.xml --multi-module app/src/main/java,lib/src/main/scala`
+1. `codacy-coverage-reporter report -r coverage.xml --multi-module app/src/main/java,lib/src/main/scala`
 
 **Other Languages**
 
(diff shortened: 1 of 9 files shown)


nenick:
If you want this pull request, than please help me to resolve the code issues. I'm not so familiar with Scala.
How to workaround the
* Found unnamed result or parameter
* Usage of get on optional type.

nenick:
Sample project use gradle and jacoco.

```
project
- app module
- lib module
- integration tests module
```

When runing the integration tests, I want a full coverage report including the files of each module. Jacoco produce this report how I expect it, but it's lacks the full path information.

> two files with same path inside different modules

Yes, multiple files with same name and package could result in strange effects. First we can add a check, which would report this unsupported situation.

ljmf00:
Hi @nenick,

This PR review led to an in depth internal discussion on whether this responsibility should be of the coverage tool or the reporter. And we think it should be with the coverage tools since they have more context and most of them are able to solve this problem in a good way.

Thank you for the time and effort you put into this PR. Although this was not the outcome you were expecting, we hope you won’t hesitate in making future contributions to improve Codacy.

PR #113, opened 2019-03-10
Native binary

rtfpessoa:
Depends on https://github.com/codacy/coverage-parser/pull/27


.circleci/config.yml
```diff
@@ -0,0 +1,170 @@
...
+            sbt ";+publishSigned;sonatypeReleaseAll"
+      - *clean_sbt_cache
+      - *save_sbt_cache
+
+  publish_binary:
+    <<: *default_sbt_job
+    steps:
+      - attach_workspace:
+          at: ~/
+      - *restore_sbt_cache
```
andreaTP: if I understand this correctly, this step doesn't require cache at all (restore nor save)

README.md
```diff
@@ -15,26 +15,33 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
...
+1. Setup the project API token. You can find the token in Project -> Settings -> Integrations -> Project API.
 
 Then set it in your terminal, replacing %Project_Token% with your own token:
 
 ```
 export CODACY_PROJECT_TOKEN=%Project_Token%
 ```
 
-3. Run the command below (changing <version> for the version you just downloaded)
+### Linux amd64
```
andreaTP: we can probably release also `darwin` binaries with circle in another iteration, I guess...

.circleci/config.yml
```diff
@@ -0,0 +1,170 @@
...
+      - run:
+          name: Publish the library
+          command: |
+            ./scripts/publish-native.sh -n codacy-coverage-reporter -m com.codacy.CodacyCoverageReporter -t docker $(cat .version)
+
+            mkdir -p /tmp/binary
+            cp ~/workdir/codacy-coverage-reporter-linux-$(cat .version) /tmp/binary/codacy-coverage-reporter-linux
+            curl -fL https://getcli.jfrog.io | sh
+            ./jfrog bt config --interactive=false --user codacy-ci --key $BINTRAY_API_KEY
```
andreaTP: just noticed that you committed 16Mb of `jfrog` CLI binaries ... 
 1 - remove the binary from the repository
 2 - evaluate to use just `curl` commands for uploading artifacts (Artifactory does support it and it's pretty easy)

scripts/publish-native.sh
```diff
@@ -0,0 +1,125 @@
...
+#   * `native` - Builds the binary for the machine arch (Requires GraalVM)
+#   * `docker` - Builds the binary for linux using a docker (Requires Docker)
+#
+# Example:
+# ./scripts/publish-native.sh -n codacy-coverage-reporter -m com.codacy.CodacyCoverageReporter -t docker 1.0.0
+#
+
+set -e
+
+VERSION="1.0.0-$(git symbolic-ref --short HEAD)-SNAPSHOT"
```
andreaTP: version comes from git-version here? can we take it as an argument

src/main/scala/com/codacy/CodacyCoverageReporter.scala
```diff
@@ -4,15 +4,16 @@ import com.codacy.configuration.parser.{CommandConfiguration, ConfigurationParsi
...
-object CodacyCoverageReporter extends ConfigurationParsingApp {
+object CodacyCoverageReporter extends ConfigurationParsingApp with StrictLogging {
 
   def run(commandConfig: CommandConfiguration): Unit = {
     val components = new Components(commandConfig)
 
     val validatedConfig = components.validatedConfig
 
-    val logger = LoggerHelper.logger(getClass, validatedConfig)
+    LoggerHelper.setLoggerLevel(logger, validatedConfig.baseConfig.debug)
```
andreaTP: why: `validatedConfig.baseConfig.debug` ?

src/main/scala/com/codacy/helpers/LoggerHelper.scala
```diff
@@ -1,30 +1,20 @@
...
-    setLoggerLevel(logger, debug)
-    logger
-  }
-
-  private def setLoggerLevel(logger: org.log4s.Logger, debug: Boolean): Unit = {
+  def setLoggerLevel(logger: Logger, debug: Boolean): Unit = {
     if (debug) {
-      logger.logger
+      logger.underlying
         .asInstanceOf[ch.qos.logback.classic.Logger]
```
andreaTP: `match` instead of casting, also avoid the repetition e.g.:
```scala
val level = if (debug) Level.DEBUG else Level.INFO
logger.underlying match {
  case l: ch.qos.logback.classic.Logger => l.setLevel(level)
  case _ =>
}
```

src/main/scala/com/codacy/rules/ConfigurationRules.scala
```diff
@@ -6,24 +6,22 @@ import cats.implicits._
...
 import scala.language.implicitConversions
 import scala.util.Try
 
-class ConfigurationRules(cmdConfig: CommandConfiguration) {
+class ConfigurationRules(cmdConfig: CommandConfiguration) extends StrictLogging {
   private val publicApiBaseUrl = "https://api.codacy.com"
 
-  private val logger = LoggerHelper.logger(getClass, cmdConfig)
-
+  LoggerHelper.setLoggerLevel(logger, cmdConfig.baseConfig.debug)
```
andreaTP: The Log level should be set only once at the start of the application, at best of my understanding, not every time you have a class that extends a Logger.

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -3,25 +3,22 @@ package com.codacy.rules
...
+import com.typesafe.scalalogging.StrictLogging
 
 import scala.util.{Failure, Success}
 
-class ReportRules(config: Configuration,
-                  coverageServices: => CoverageServices) {
+class ReportRules(config: Configuration, coverageServices: => CoverageServices) extends StrictLogging {
 
-  private val logger: Logger = LoggerHelper.logger(getClass, config)
+  LoggerHelper.setLoggerLevel(logger, config.baseConfig.debug)
```
andreaTP: the comments above.

README.md
rtfpessoa: I think you have to pay. Ideally we can release a docker using scratch base latter.

.circleci/config.yml
rtfpessoa: it requires the compiled classes

.circleci/config.yml
rtfpessoa: This is already working and is just a command to install. (also being the recommended way)

scripts/publish-native.sh
rtfpessoa: This is just the default. It is being received as an argument.

src/main/scala/com/codacy/CodacyCoverageReporter.scala
rtfpessoa: What would be the alternative?

src/main/scala/com/codacy/helpers/LoggerHelper.scala
rtfpessoa: Nice catch.

src/main/scala/com/codacy/rules/ConfigurationRules.scala
rtfpessoa: I was not sure about that.

PR #123, opened 2019-07-22
Change image to official GraalVM

bmbferreira:
The existing image tag on the script was removed and the script doesn't run anymore when we're targeting docker.

### DIFF
--- a/scripts/publish-native.sh
+++ b/scripts/publish-native.sh
@@ -115,8 +115,8 @@ case "$TARGET" in
       -v $HOME/.ivy2:$HOME/.ivy2 \
       -v $HOME/.sbt:$HOME/.sbt \
       -v $PWD:$PWD \
-      oracle/graalvm-ce:1.0.0-rc14 \
-        -c "${BUILD_CMD}"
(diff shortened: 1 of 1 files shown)


rtfpessoa:
The title is very misleading. I think something like `Change image to official GraalVM` would be more appropriate.

scripts/publish-native.sh
```diff
@@ -115,8 +115,8 @@ case "$TARGET" in
       -v $HOME/.ivy2:$HOME/.ivy2 \
       -v $HOME/.sbt:$HOME/.sbt \
       -v $PWD:$PWD \
-      oracle/graalvm-ce:1.0.0-rc14 \
-        -c "${BUILD_CMD}"
+      oracle/graalvm-ce:19.1.1 \
+        -c "gu install native-image && ${BUILD_CMD}"
```
andreaTP: this is fine as a fix, it will be nice to have it bundled into the image as a further improvement.

andreaTP:
as a side note, since the previous image got removed from Dockerhub I think it's safer if in the future we start publishing our own images based on upstream.

PR #126, opened 2019-07-24
fix: Adds new flags to fix the native image build (fixes #125)

### LINKED ISSUE #125: 6.0.1 - Could not find or load main class com.codacy.CodacyCoverageReporter
Using the following line to run codacy and upload coverage reports
```
curl -Ls -o codacy-coverage-reporter "$(curl -Ls https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/latest | jq -r '.assets | map({name, browser_download_url} | select(.name | contains("codacy-coverage-reporter-linux"))) | .[0].browser_download_url')"
chmod +x codacy-coverage-reporter
./codacy-coverage-reporter report -l Java -r build/reports/jacoco/test/jacocoTestReport.xml
```
It gives me an error with the latest version 6.0.1
```
Error: Could not find or load main class com.codacy.CodacyCoverageReporter
Caused by: java.lang.ClassNotFoundException: com.codacy.CodacyCoverageReporter
Exited with code 1
```

If I pin it back to 6.0.0, it works without issue.

bmbferreira:
The native image build got broken after upgrading the graalvm image.
We had to add a new flag to initialize classes and implicitly all of their superclasses during image building (--initialize-at-build-time).
The --no-fallback flag is to avoid falling back to JDK if the build is not successfull and the -report-unsupported-elements-at-runtime flag is to report usage of unsupported methods and fields at run time when they are accessed the first time, instead of as an error during image building.

### DIFF
--- a/scripts/publish-native.sh
+++ b/scripts/publish-native.sh
@@ -81,6 +81,8 @@ function build_cmd() {
   FLAGS+=' --enable-url-protocols=http,https,file,jar --enable-all-security-services'
   FLAGS+=' -H:+JNI -H:IncludeResourceBundles=com.sun.org.apache.xerces.internal.impl.msg.XMLMessages'
   FLAGS+=' -H:+ReportExceptionStackTraces'
+  FLAGS+=' --no-fallback --initialize-at-build-time'
+  FLAGS+=' --report-unsupported-elements-at-runtime'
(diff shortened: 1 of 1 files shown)


bmbferreira:
Fixes #125

PR #133, opened 2019-09-09
Add a safe check for coverage report token

ljmf00:


.circleci/config.yml
```diff
@@ -91,7 +91,12 @@ jobs:
       - run:
           name: Coverage
           command: |
-            sbt clean compile coverage test coverageReport coverageAggregate codacyCoverage
+            if [ -z ${CODACY_PROJECT_TOKEN+x} ]; then
```
lolgab: In this way we could forgot to add the CODACY_PROJECT_TOKEN because the CI fails? Isn't better having the coverage always running?

ljmf00: No. Because the project token should be a private token, and therefore shouldn't be available for PRs from outside contributors (on circle CI, for security reasons). To make the CI pass, you should have this safe check, otherwise, every PR from a fork will fail.

PR #146, opened 2019-11-26
Fix SSL Hanshake and DNS issues

andreaTP:
Here we propose to bump GraalVM to 19.3.0 to properly include `libsunec.so`.
There is known bug in GraalVM and DNS resolution( https://github.com/oracle/graal/issues/624 ), so here I propose a workaround.

Please @ljmf00 doublecheck and take over this.


ljmf00:
I tested on my machine, and unfortunately it only works if we change the `/etc/hosts`. The user should be aware that this is an hacky workaround . I also found a way to substitute the host instead of just appending:

 `sed '/api.codacy.com/d' /etc/hosts | (cat && echo "$(dig +short api.codacy.com | tail -n1) api.codacy.com") >> /etc/hosts`

ljmf00:
@andreaTP check my patch and it's good to go

andreaTP:
@ljmf00 since I'm the author of the PR I cannot approve :-) good to go for me

Issue #160, opened 2020-02-24
Make script use native-image when possible


lolgab:
codacy-coverage-reporter script use java by default. This is all but ideal for non-java users!
We should make it use the native-image when it's possible (linux machine), using java as a fallback.

PR #163, opened 2020-02-28
feature: get.sh script get native version when on linux distro

josemiguelmelo:
https://github.com/codacy/codacy-coverage-reporter/issues/160

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -1,4 +1,9 @@
-#!/usr/bin/env bash
+#!/bin/sh
+path_to_bash=$(which bash)
+is_running_in_bash=$(ps aux | grep $0 | grep bash | wc -l)
+if [ -x "$path_to_bash" ] && [ $is_running_in_bash = 0 ]; then
+    exec bash "$0" "$@";
+fi
 
 set -e +o pipefail
 
@@ -101,7 +106,7 @@ codacy_reporter_jar_start_cmd() {
 
 run_command=""
 
-if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+if [ "$(expr substr $(uname -s) 1 5)" = "Linux" ]; then
     codacy_reporter_native_start_cmd
 else
     codacy_reporter_jar_start_cmd
--- a/get.sh
+++ b/get.sh
@@ -1,9 +1,4 @@
-#!/bin/sh
-path_to_bash=$(which bash)
-is_running_in_bash=$(ps aux | grep $0 | grep bash | wc -l)
-if [ -x "$path_to_bash" ] && [ $is_running_in_bash = 0 ]; then
-    exec bash "$0" "$@";
-fi
+#!/usr/bin/env bash
 
 set -e +o pipefail
 
@@ -82,7 +77,7 @@ download_coverage_reporter() {
     if [ ! -f "$codacy_reporter" ]
     then
         log "$i" "Download the codacy reporter $1... ($CODACY_REPORTER_VERSION)"
-        curl -LS -o "$codacy_reporter" "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION | grep browser_download_url | grep $1 | cut -d '"' -f 4)"
+        curl -# -LS -o "$codacy_reporter" "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION | grep browser_download_url | grep $1 | cut -d '"' -f 4)"
     else
         log "$i" "Using codacy reporter $1 from cache"
     fi
@@ -95,6 +90,7 @@ run() {
 codacy_reporter_native_start_cmd() {
     codacy_reporter="$codacy_temp_folder/codacy-coverage-reporter"    
     download_coverage_reporter "linux"
+    chmod +x $codacy_reporter
     run_command="$codacy_reporter"
 }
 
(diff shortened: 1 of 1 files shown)


get.sh
```diff
@@ -1,4 +1,9 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
ljmf00: Take this as a tip: I want to mention some info about why I used `/usr/bin/env` on the shebang. To make the script more portable and make it searchable on the `$PATH` variable, it could be useful using it instead of hardcoding the binary path. For example, it could be useful for systems like FreeBSD, OpenBSD or other system that doesn't unsure you that the binaries are not stored on `/bin`. In this case it doesn't happen because `sh` is different from `bash` on most of the systems, so it's generally fine to do `#!/bin/sh`.

More info about that if you are interested: https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

Also, on the lines below, you are rerunning the script using bash (without exiting tho). If your script is fully SH complaint, running with `/bin/sh` is fine, making running it under bash useless and more restrictive. But if your script is only BASH complaint, which is when you deal with some cool bash features (called bashisms), you shouldn't run it using only SH complaint shell. I will consider removing this wrapper, which is also possibly dangerous by assuming only one process with this name, and instead rely on the PID.

TL;DR;

If you want to make this script truly sh compatible, you could use some tools to help porting it: https://linux.die.net/man/1/checkbashisms . Also consider using `shellcheck`.

Nevertheless, continue with the amazing work guys! <3

lolgab: Thank you for the hints Luís ❤️.
We did this because we want seamless compatibility in the two most used Linux distributions for CI: Alpine and Ubuntu.
From one end Alpine doesn't have `bash` at all. From the other Ubuntu has `dash` as default `sh` shell but has bash preinstalled.
But, as fas as we know, `dash` doesn't support pipefail out of the box. So the idea is to run with sh but fallback to bash if it's installed on the system.

get.sh
```diff
@@ -73,19 +78,48 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+
+codacy_reporter_jar_start_cmd() {
+    codacy_reporter="$codacy_temp_folder/codacy-coverage-reporter-assembly.jar"
+    download_coverage_reporter "jar"
+    run_command="java -jar \"$codacy_reporter\""
+}
+
+run_command=""
+
+if [ "$(expr substr $(uname -s) 1 5)" = "Linux" ]; then
```
mrfyda: ☝️this causes a `expr: syntax error` on Mac

get.sh
```diff
@@ -73,19 +78,48 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+}
 
-if [ ! -f "$codacy_reporter" ]
-then
-    log "$i" "Download the codacy reporter... ($CODACY_REPORTER_VERSION)"
-    curl -LS -o "$codacy_reporter" "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION | grep browser_download_url | grep jar | cut -d '"' -f 4)"
+codacy_reporter_native_start_cmd() {
+    codacy_reporter="$codacy_temp_folder/codacy-coverage-reporter"    
+    download_coverage_reporter "linux"
+    run_command="$codacy_reporter"
```
mrfyda: ☝️this fails with `/dev/fd/63: line 92: .codacy-coverage/codacy-coverage-reporter: Permission denied`
after `chmod +x codacy-coverage-reporter` it is successful

get.sh
josemiguelmelo: Replaced expr with `uname` to fix it 👍

get.sh
josemiguelmelo: Added `chmod +x $codacy_reporter` to fix it 👍

get.sh
ljmf00: Sure! About the mechanism to detect `bash`, this could be a better approach, using PID instead of the script name, as the PID will be unique, in this case. :wink: 

```sh
ps -C bash -o pid= | grep "$$"
```
Also, you can check the running shell with `ps` using the PID too.

```sh
ps -o command= "$$"
```
Or even, instead of the entire path, only the command name:

```sh
ps -o comm= "$$"
```

get.sh
```diff
@@ -73,19 +74,49 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
     CODACY_REPORTER_VERSION="latest"
 fi
 
-codacy_reporter="$codacy_temp_folder/codacy-coverage-reporter-assembly.jar"
+download_coverage_reporter() {
+    if [ ! -f "$codacy_reporter" ]
+    then
+        log "$i" "Download the codacy reporter $1... ($CODACY_REPORTER_VERSION)"
+        curl -LS -o "$codacy_reporter" "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION | grep browser_download_url | grep $1 | cut -d '"' -f 4)"
```
mrfyda:
```suggestion
        curl -# -LS -o "$codacy_reporter" "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION | grep browser_download_url | grep $1 | cut -d '"' -f 4)"
```
this will output a one line progress bar `###...## 100.0%`

get.sh
mrfyda: it worked 🚀 
👉 https://github.com/mrfyda/codacy-coverage-reporter-action/runs/478094531?check_suite_focus=true#step:4:12

get.sh
josemiguelmelo: Really nice suggestion, it looks a lot better with the progress bar 🙏👍

get.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+
```
lolgab: Remove this extra space here

PR #177, opened 2020-03-11
docs: Update NotFound error message

mrfyda:
The token can now be either a project or account token


src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -234,7 +234,7 @@ class ReportRules(coverageServices: => CoverageServices) extends StrictLogging {
     */
   private[rules] def handleFailedResponse(response: FailedResponse): String = {
     if (response.message.contains("not found")) {
-      """Request URL not found. (Check if the project token or the API base URL are valid)"""
+      """Request URL not found. (Check if the token you are using or the API base URL are valid)"""
```
lolgab: You can use a single quoted string:
```suggestion
      "Request URL not found. (Check if the token you are using or the API base URL are valid)"
```

PR #178, opened 2020-03-11
clean: Use sbt-native-packager to build native-image

josemiguelmelo:


build.sbt
```diff
@@ -50,3 +50,19 @@ scmInfo := Some(
 
 fork in Test := true
 cancelable in Global := true
+
+enablePlugins(GraalVMNativeImagePlugin)
+graalVMNativeImageGraalVersion := Some("19.3.0-java8")
```
lolgab: Update graalvm here

.circleci/config.yml
```diff
@@ -94,26 +108,23 @@ workflows:
...
           cmd: |
-            sbt "assembly"
-
-            ./scripts/publish-native.sh -n codacy-coverage-reporter -m com.codacy.CodacyCoverageReporter -t docker $(cat .version)
-
-            mkdir -p ~/workdir/artifacts/
-            mv codacy-coverage-reporter-linux-$(cat .version) ~/workdir/artifacts/
-            mv ~/workdir/target/codacy-coverage-reporter-assembly-$(cat .version).jar ~/workdir/artifacts/
+            sbt assembly
+            sbt 'show graalvm-native-image:packageBin'
```
lolgab:
```suggestion
            sbt graalvm-native-image:packageBin
```

lolgab: I don't think you need assembly

lolgab: If memory is enough you don't have to pay sbt startup time twice

build.sbt
josemiguelmelo: ✅

.circleci/config.yml
```diff
@@ -57,6 +57,20 @@ jobs:
...
+  package_artifacts:
+    docker:
+      - image: hairyhenderson/upx:latest
+    working_directory: ~/workdir
+    steps:
+      - run: apk add ca-certificates
+      - attach_workspace:
+          at: ~/workdir
+      - run: |
+          upx --lzma -o codacy-coverage-reporter-compressed ~/workdir/target/graalvm-native-image/codacy-coverage-reporter
```
andreaTP:
```
mkdir -p ~/workdir/artifacts/
cp ~/workdir/target/codacy-coverage-reporter-assembly-$(cat .version).jar ~/workdir/artifacts/
upx --lzma -o codacy-coverage-reporter-compressed ~/workdir/artifacts/codacy-coverage-reporter-linux-$(cat .version)
```

WDYT?

.circleci/config.yml
```diff
@@ -94,26 +108,23 @@ workflows:
...
+          name: create_artifacts
           cmd: |
-            sbt "assembly"
-
-            ./scripts/publish-native.sh -n codacy-coverage-reporter -m com.codacy.CodacyCoverageReporter -t docker $(cat .version)
-
-            mkdir -p ~/workdir/artifacts/
-            mv codacy-coverage-reporter-linux-$(cat .version) ~/workdir/artifacts/
-            mv ~/workdir/target/codacy-coverage-reporter-assembly-$(cat .version).jar ~/workdir/artifacts/
+            sbt assembly
```
andreaTP: you can squash the two commands into one single sbt invocation.

.circleci/config.yml
```diff
@@ -57,6 +57,20 @@ jobs:
           name: Publish final orb
           command: circleci orb publish orb.yml codacy/coverage-reporter@$(cat .version) --token $CIRCLE_TOKEN
 
+  package_artifacts:
+    docker:
+      - image: hairyhenderson/upx:latest
```
andreaTP: please avoid using moving tags.

.circleci/config.yml
josemiguelmelo: Nice suggestion, it worked 🙏

.circleci/config.yml
josemiguelmelo: Done ✅

.circleci/config.yml
josemiguelmelo: You're right! Set a specific image version ✅

lolgab:
Next is to port this to every GraalVM binary we have!

PR #182, opened 2020-03-19
test: Add integration tests for get.sh script

franciscodua:
Add integration test for Alpine based systems
Add integration test for Ubuntu based systems

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,8 @@ commands:
             export CODACY_TOOL_PATH=$TEST_CODACY_TOOL_PATH
             set +e
             << parameters.executor >> get.sh report --commit-uuid $TEST_COMMIT_UUID -r $TEST_CODACY_REPORT_PATH --codacy-api-base-url http://localhost:1080
-            if [ $? -ne << parameters.error_code >> ]; then exit 1; fi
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -74,6 +74,55 @@ jobs:
...
+          command: timeout 60 bash -c 'while [[ "$(curl -X PUT -o /dev/null -w ''%{http_code}'' localhost:1080/mockserver/status)" != "200" ]]; do sleep 2; done' || false
+      - run:
+          name: configure mock server
+          command: curl -v -X PUT localhost:1080/mockserver/expectation -d "$(cat integration-tests/mock-server-config.json)"
+      - run:
+          name: test coverage reporting
+          command: |
+            export CODACY_PROJECT_TOKEN=$TEST_CODACY_PROJECT_TOKEN
+            export CODACY_TOOL_PATH=$TEST_CODACY_TOOL_PATH
+            bash get.sh report --commit-uuid $TEST_COMMIT_UUID -r $TEST_CODACY_REPORT_PATH --codacy-api-base-url http://localhost:1080
```
josemiguelmelo: Instead of having this duplicated, you could try to create a command (https://circleci.com/docs/2.0/reusing-config/#the-commands-key). What do you think?

franciscodua: Done. Way better now

PR #188, opened 2020-03-26
Fixes to the IT tests

andreaTP:
Three commits here for the 3 fixes :-)

PR #189, opened 2020-03-26
Move back to fetch binaries from Bintray

andreaTP:


get.sh
```diff
@@ -73,18 +73,36 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+            CODACY_REPORTER_VERSION="$(wget -O - $1 | download_url)"
+        else
+            fail_curl_wget
+        fi
+    fi
+}
+
 download_url() {
-    grep browser_download_url | grep $1 | cut -d '"' -f 4
+    sed -e 's/.*name.*\([0-9]\+[.][0-9]\+[.][0-9]\+\).*/\1/'
```
andreaTP: this will only match proper releases, IMHO is a little limitation (and is probably a stronger command than before)

get.sh
```diff
@@ -73,18 +73,36 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
 }
 
 download_using_wget_or_curl() {
-    api_url="https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION"
+    buntray_latest_api_url="https://api.bintray.com/packages/codacy/Binaries/codacy-coverage-reporter/versions/_latest"
```
franciscodua: minor typo
```suggestion
    bintray_latest_api_url="https://api.bintray.com/packages/codacy/Binaries/codacy-coverage-reporter/versions/_latest"
```

andreaTP: :man_facepalming:

andreaTP: fixed and squashed the commits

andreaTP:
ok, this broke because I published from a branch to bintray.
The CI will be fixed merging:
https://github.com/codacy/codacy-coverage-reporter/pull/188
first

get.sh
```diff
@@ -73,18 +73,36 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
-    grep browser_download_url | grep $1 | cut -d '"' -f 4
+    sed -e 's/.*name.*\([0-9]\+[.][0-9]\+[.][0-9]\+\).*/\1/'
 }
 
 download_using_wget_or_curl() {
-    api_url="https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION"
+    bintray_latest_api_url="https://api.bintray.com/packages/codacy/Binaries/codacy-coverage-reporter/versions/_latest"
+    get_version $bintray_latest_api_url $1
+    bintray_api_url="https://dl.bintray.com/codacy/Binaries/$CODACY_REPORTER_VERSION/codacy-coverage-reporter-$1"
     if [ -x "$(which curl)" ]; then
```
rtfpessoa: We should not have to do this if every time. Make downloading a function

get.sh
```diff
@@ -73,18 +73,36 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
 fi
 
+fail_curl_wget() {
+    fatal "Could not find curl or wget, please install one."
+}
+
+get_version() {
+    if [ "$CODACY_REPORTER_VERSION" == "latest" ]; then
+        if [ -x "$(which curl)" ]; then
+            CODACY_REPORTER_VERSION="$(curl -LSs $1 | download_url)"
```
rtfpessoa: Please do not use global variables. Functions have inputs and an output. Use that output to return the value.

get.sh
rtfpessoa: This is only used in get_version and is used in all scenarios.
Do not refactor this, pipe it in the end before returning.

get.sh
```diff
@@ -73,18 +73,36 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+}
+
 download_url() {
-    grep browser_download_url | grep $1 | cut -d '"' -f 4
+    sed -e 's/.*name.*\([0-9]\+[.][0-9]\+[.][0-9]\+\).*/\1/'
 }
 
 download_using_wget_or_curl() {
-    api_url="https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/$CODACY_REPORTER_VERSION"
+    bintray_latest_api_url="https://api.bintray.com/packages/codacy/Binaries/codacy-coverage-reporter/versions/_latest"
```
rtfpessoa: This part is very confusing after refactoring the previous methods this should get easier to follow

get.sh
```diff
@@ -73,21 +73,32 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+        echo "wget -O"
     else
         fatal "Could not find curl or wget, please install one."
     fi
 }
 
+get_version() {
+    if [ "$CODACY_REPORTER_VERSION" == "latest" ]; then
+        bintray_latest_api_url="https://api.bintray.com/packages/codacy/Binaries/codacy-coverage-reporter/versions/_latest"
+        latest="$($(curl_or_wget) $bintray_latest_api_url | sed -e 's/.*name.*\([0-9]\+[.][0-9]\+[.][0-9]\+\).*/\1/')"
```
rtfpessoa: after the refactor this line should be enough
```suggestion
        latest="$(download $bintray_latest_api_url | sed -e 's/.*name.*\([0-9]\+[.][0-9]\+[.][0-9]\+\).*/\1/')"
```

mrfyda:
@andreaTP did you consider gh packages for this? https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-apache-maven-for-use-with-github-packages

get.sh
```diff
@@ -73,21 +73,32 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+        echo $latest
+    else
+        echo "$CODACY_REPORTER_VERSION"
+    fi
+}
+
+download_using_wget_or_curl() {
+    bintray_api_url="https://dl.bintray.com/codacy/Binaries/$(get_version)/codacy-coverage-reporter-$1"
```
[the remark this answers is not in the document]
andreaTP: done the `local` I'm not going to refactor further this file in this PR since it's out of the scope

andreaTP:
@mrfyda here we are trying to go back to use bintray for the distribution(as it was before) instead of using GH releases simply because GH is going to Rate Limit.
Also we are publishing and re-using jar packages as well as native executables.

mrfyda:
was just curious. but gh packages is not the same as gh releases

get.sh
```diff
@@ -73,21 +73,42 @@ if [ -z "$CODACY_REPORTER_VERSION" ]; then
...
+        echo $latest
+    else
+        echo "$CODACY_REPORTER_VERSION"
+    fi
+}
+
+download_using_wget_or_curl() {
+    local architecture=$1
+    bintray_api_url="https://dl.bintray.com/codacy/Binaries/$(get_version)/codacy-coverage-reporter-$architecture"
+    $(download "$bintray_api_url" "$codacy_reporter")
```
rtfpessoa: Why are you invoking the output?

andreaTP: removed

get.sh
```diff
@@ -104,14 +116,14 @@ run() {
 
 codacy_reporter_native_start_cmd() {
     codacy_reporter="$codacy_temp_folder/codacy-coverage-reporter"
```
rtfpessoa: I really think this global variable is a big issue here.

andreaTP: agree, haven't noticed that it was polluting everything, made local

PR #198, opened 2020-04-16
doc: Add version restriction to PHPUnit xml

franciscodua:
PHPUnit xml format changed after version 4. And this tool does not support newer versions.
To use this tool with newer versions, users should use the flag `--coverage-clover` and send the report using the clover format.


mrfyda:
> To use this tool with newer versions, users should use the flag --coverage-clover and send the report using the clover format.

can we add this to the troubleshooting section? 🤓

PR #200, opened 2020-04-17
Change docs to use mkdocs

mrfyda:
This way they can be included with the product docs: https://docs.dev.codacy.org/coverage-reporter/


README.md
```diff
@@ -9,182 +9,7 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
...
-echo "$(dig +short api.codacy.com | tail -n1) api.codacy.com" >> /etc/hosts
-```
-
-### `coverage-xml/index.xml generated an empty result`
-
-If you are using PHPUnit version 5 or above to generate your coverage report, you must output the report using the clover format. The codacy-coverage-reporter supports the PHPUnit xml format only for versions equal or lower than 4.
-You can change the output format by replacing the `--coverage-xml <dir>` flag by `--coverage-clover <file>`.
-
-For more information on the PHPUnit command line check [here](https://phpunit.readthedocs.io/en/9.1/textui.html).
+[Get started](http://docs.dev.codacy.org/coverage-reporter)
```
mrfyda, line 12: @pauloribeiro-codacy since this is the live version of the docs we should maybe point to the docs folder of this repo instead

pauloribeiro-codacy: Yes, I think it's better for now, as the docs under http://docs.dev.codacy.org/ are not quite ready for consumption yet. :)

The MarkDown in this repository will not render 100% on GitHub, but even so I think it's best to have:

```suggestion
[Get started](https://github.com/codacy/codacy-coverage-reporter/blob/master/docs/index.md)
```

mrfyda:
😀

mrfyda:
- [ ] After merging, we should make `ci/circleci: build_mkdocs` required as well

PR #230, opened 2020-06-09
fix: Send final report in automatic mode :breaking:

lolgab:
Send final report when guessing multiple files and the --partial
flag is not specified.
In this situation the previous behaviour was to send some partial
reports but without a final at the end. This way users need to
send a manual final right after.
With this change the final is sent automatically.

### DIFF
--- a/src/main/scala/com/codacy/rules/ReportRules.scala
+++ b/src/main/scala/com/codacy/rules/ReportRules.scala
@@ -38,7 +38,7 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
       val filesEither = guessReportFiles(config.coverageReports, rootProjectDirIterator)
 
       filesEither.flatMap { files =>
-        def sendFilesReport(partial: Boolean) = {
+        def sendFilesReport(partial: Boolean = config.partial) = {
           val finalConfig = config.copy(partial = partial)
           files
             .map { file =>
@@ -63,7 +63,7 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
             f <- finalReport(FinalConfig(config.baseConfig))
           } yield f
         } else {
-          sendFilesReport(partial = true)
+          sendFilesReport()
         }
       }
     }
--- a/src/main/scala/com/codacy/rules/ReportRules.scala
+++ b/src/main/scala/com/codacy/rules/ReportRules.scala
@@ -38,26 +38,33 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
       val filesEither = guessReportFiles(config.coverageReports, rootProjectDirIterator)
 
       filesEither.flatMap { files =>
-        val finalConfig = if (files.length > 1 && !config.partial) {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -38,26 +38,33 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
...
+              } yield { success }
+            }
+            .collectFirst {
+              case Left(l) => Left(l)
+            }
+            .getOrElse(Right("All coverage data uploaded."))
+        }
+        if (files.length > 1 && !config.partial) {
           logger.info("More than one file. Considering a partial report")
-          config.copy(partial = true)
```
franciscodua, line 43: Isn't it enough to skip this? Basically, if we didn't specify partial and we have multiple reports, it's not partial. Then, in the send the report we send a final if needed

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -38,26 +38,33 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
...
-            case Left(l) => Left(l)
-          }
-          .getOrElse(Right("All coverage data uploaded."))
+          for {
+            _ <- sendFilesReport(partial = true)
+            f <- finalReport(FinalConfig(config.baseConfig))
+          } yield f
+        } else {
+          sendFilesReport(partial = true)
+        }
```
josemiguelmelo, line 67: You could simply add the if at the end to send the `finalReport`. Something like this:

```suggestion
   sendFilesReport(partial = true)
   if (files.length > 1 && !config.partial) {
       finalReport(FinalConfig(config.baseConfig))
    } else {
       sendFilesReport(partial = true)
    }
```

lolgab: The code was wrong.
In the normal case I should call `sendReport` passing the original partial setting.
And in any case you need the for comprehension because you don't want to pass the `final` if the partial send was not successful.

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -38,26 +38,33 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
       val filesEither = guessReportFiles(config.coverageReports, rootProjectDirIterator)
 
       filesEither.flatMap { files =>
-        val finalConfig = if (files.length > 1 && !config.partial) {
+        def sendFilesReport(partial: Boolean = config.partial) = {
```
franciscodua, line 41: Can we make this a private class method?
Possibly just me, but I find it harder to read now. The def inside a flatMap, inside a method.

Just food for thought...

lolgab: Pushed a new commit with it as a private method.
I needed, however, to pass some more parameters.
Your call, if you prefer the old version feel free to `git reset --hard HEAD~1` and merge as it was before.

PR #244, opened 2020-08-26
doc: Refactor documentation

pauloribeiro-codacy:
I'm refactoring and simplifying the current documentation structure around coverage. See https://github.com/codacy/docs/issues/84 for more details.

The documentation from this repository is included in the documentation generated from codacy/docs: https://docs.codacy.com/coverage-reporter/


pauloribeiro-codacy:
We should also be able to list the language aliases that the option `--language` accepts.

pauloribeiro-codacy:
Let's also clarify that Codacy Coverage Reporter supports a list of **coverage report formats**. The tools that generate these report formats (and the languages associated with the tools) should be in a different page dedicated to helping the users set up coverage. This will also simplify the main procedure on how to configure Codacy Coverage Reporter to upload the coverage data.

docs/troubleshooting-common-issues.md
```diff
@@ -0,0 +1,76 @@
...
+## Submitting coverage for unsupported languages
+
+If your language is not in the list of supported languages, you can still send coverage to Codacy. You can do it by providing the correct `--language` name and adding the `--force-language` flag. For example:
+
+```bash
+bash <(curl -Ls https://coverage.codacy.com/get.sh) report
+  --language Kotlin --force-language
+```
+
+See the [list of languages](https://github.com/codacy/codacy-plugins-api/blob/master/src/main/scala/com/codacy/plugins/api/languages/Language.scala#L43) that you can specify using the flag `--language`.
```
pauloribeiro-codacy, line 14: This is not ideal since the target file can change and we no longer point to the correct line.

docs/adding-coverage-to-your-repository.md
```diff
@@ -0,0 +1,121 @@
...
+
+| Report formats                                              | Report file names            |
+| ----------------------------------------------------------- | ---------------------------- |
+| [Clover](https://openclover.org/)                           | clover.xml                   |
+| [Cobertura](http://cobertura.github.io/cobertura/)          | cobertura.xml                |
+| [dotCover](https://www.jetbrains.com/dotcover/) XML         | dotcover.xml                 |
+| [JaCoCo](https://www.jacoco.org/)                           | jacoco\*.xml                 |
+| [LCOV](http://ltp.sourceforge.net/coverage/lcov/readme.php) | lcov.info, lcov.dat, \*.lcov |
+| [OpenCover](https://github.com/OpenCover/opencover)         | opencover.xml                |
+| [PHPUnit](https://phpunit.de/) XML (version &lt;= 4)        | coverage-xml/index.xml       |
```
pauloribeiro-codacy, line 15: The links in this table are mostly for the tools that generate each report format and not for information on the report format itself. I couldn't find specific information for most of the formats. Maybe it's better not to include the links? :thinking:

docs/generating-coverage-reports.md
```diff
@@ -0,0 +1,35 @@
+# Generating coverage reports
```
pauloribeiro-codacy, line 1: The help provided in this page is quite basic for now. We can think about improving the information in this section if we understand that a large number of customers need help with generating the coverage reports to start with.

pauloribeiro-codacy:
To help with the pull request review, here are links to the previews of the complete set of documentation pages:

* [Generating coverage reports](https://github.com/codacy/codacy-coverage-reporter/blob/36b53b35f8f54fac69b3c1a77ae8350afab2bda2/docs/generating-coverage-reports.md)
* [Adding coverage to your repository](https://github.com/codacy/codacy-coverage-reporter/blob/36b53b35f8f54fac69b3c1a77ae8350afab2bda2/docs/adding-coverage-to-your-repository.md)
* [Alternative ways of running Coverage Reporter](https://github.com/codacy/codacy-coverage-reporter/blob/36b53b35f8f54fac69b3c1a77ae8350afab2bda2/docs/alternative-ways-of-running-coverage-reporter.md)
* [Troubleshooting common issues](https://github.com/codacy/codacy-coverage-reporter/blob/36b53b35f8f54fac69b3c1a77ae8350afab2bda2/docs/troubleshooting-common-issues.md)

docs/adding-coverage-to-your-repository.md
pauloribeiro-codacy: Also, the information is repeated here: https://github.com/codacy/coverage-parser#supported-formats

I think it's best to have the information on the codacy/coverage-parser repository as the single source of truth. But we should simplify the table so it focuses on the actual **report formats** and not on the tools that generate them.

franciscodua: > The links in this table are mostly for the tools that generate each report format and not for information on the report format itself

True. And the name should suffice if the user wants to look it up. Do we have a list of tools that output this report here?

> I think it's best to have the information on the codacy/coverage-parser repository as the single source of truth

At the time we moved it here to avoid the indirection. People looking for this information had to come here and somehow find a link to a different repository with this information. Which might not be the best experience.

docs/generating-coverage-reports.md
franciscodua: The most valuable part is probably giving examples of tools that people can use to get coverage for their projects (and the swift workaround)

docs/troubleshooting-common-issues.md
franciscodua: Yup. But I'm not sure how we could solve this.
An alternative would be to use this page https://docs.codacy.com/getting-started/supported-languages/
But the names of the languages are not the same (example: CSharp is C# in that page).

Maybe we should have a resource with that list of languages 🤔

franciscodua (dismissed):

docs/troubleshooting-common-issues.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Submitting coverage generated by unsupported tools {: id="unsupported-tools"}
+
+If you are generating a report format that Codacy does not yet support, try using community projects such as [schrej/godacov](https://github.com/schrej/godacov) and [danielpalme/ReportGenerator](https://github.com/danielpalme/ReportGenerator), or alternatively contribute to our [codacy/coverage-parser](https://github.com/codacy/coverage-parser) project.
```
mrfyda, line 18: I think it would be more helpful to guide users towards format converters than implementations of the reporter. for example, https://github.com/t-yuki/gocover-cobertura instead of https://github.com/schrej/godacov

we can always provide better support, faster access to new features if users are using our reporter

docs/generating-coverage-reports.md
mrfyda: > The most valuable part is probably giving examples of tools that people can use to get coverage for their projects (and the swift workaround)

Yes. Ideally, for each language we would have an example of what it takes to set up coverage and report it to Codacy. See here an example: https://github.com/codecov/example-scala

docs/adding-coverage-to-your-repository.md
pauloribeiro-codacy: > > The links in this table are mostly for the tools that generate each report format and not for information on the report format itself
> 
> True. And the name should suffice if the user wants to look it up. Do we have a list of tools that output this report here?

Nice, makes sense then. The suggested tools are now on a dedicated page on how to generate coverage: [Generating coverage reports](https://github.com/codacy/codacy-coverage-reporter/blob/36b53b35f8f54fac69b3c1a77ae8350afab2bda2/docs/generating-coverage-reports.md).

> > I think it's best to have the information on the codacy/coverage-parser repository as the single source of truth
> 
> At the time we moved it here to avoid the indirection. People looking for this information had to come here and somehow find a link to a different repository with this information. Which might not be the best experience.

I agree that in terms of experience it's best to have the information streamlined here. :+1: I will open a pull request on codacy/coverage-parser to suggest removing the repeated information from there and add a pointer to the documentation instead. Hopefully this will help remind people working on that project that changes must be reflected on this documentation.

docs/troubleshooting-common-issues.md
pauloribeiro-codacy: This makes perfect sense. When I first read this information, I was convinced that the schrej/godacov project was a converter, and was surprised to see it was a reimplementation of our reporter tool. :+1:

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -1,55 +1,44 @@
...
 
-Thanks to the amazing job of [halkeye](https://github.com/halkeye) you can now send your coverage to Codacy using a [maven plugin](https://github.com/halkeye/codacy-maven-plugin)!
-
-Just follow the [instructions on his repository](https://github.com/halkeye/codacy-maven-plugin/blob/master/README.md#usage).
+Thanks to the amazing job of [Gavin Mogan](https://github.com/halkeye) you can now send your coverage to Codacy using his Maven plugin [halkeye/codacy-maven-plugin](https://github.com/halkeye/codacy-maven-plugin)! Be sure to follow the instructions on his repository.
 
 ### Travis CI
 
-If you want to use codacy with Travis CI and report coverage generated from your tests run in Travis, update your .travis.yml to include the following blocks:
+If you are using Travis CI to report coverage, update your file `.travis.yml` to include the following blocks:
```
mrfyda, line 41: Looks like this section could be updated

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -123,70 +110,71 @@ task sendCoverageToCodacy(type: JavaExec, dependsOn: jacocoTestReport) {
...
 ## Manually downloading the Java binary
 
-- Linux x86, MacOS, Windows, ...
+Use the Java binary to run Codacy Coverage reporter on other platforms, such as Linux x86, MacOS, Windows, etc.
 
-Download the latest jar and use it to post the coverage to Codacy
+You can use the scripts below to automatically check for the latest version of the Java binaries, download the binaries from either Bintray or GitHub, and run them.
 
-#### Bintray
+Using Bintray:
```
mrfyda, line 142: Is there a distinction between Bintray and GitHub that makes it relevant for users? Should we make it a choice for them?

franciscodua: That's a good question. We changed to bintray due to rate limiting on Github if I'm not mistaken. @josemiguelmelo do you know if this was the case? If it was maybe we can drop the Github part

docs/adding-coverage-to-your-repository.md
mrfyda: Keep in mind that users will be most familiar with their languages, build system, and testing frameworks; not necessarily the coverage format that combination outputs. "I'm using Java, Maven and jUnit" - ideally, this would be enough to understand if the reporter is supported.

docs/troubleshooting-common-issues.md
pauloribeiro-codacy: Yes, that could be a more elegant solution.

By the way, last week we noticed that this list of languages is repeated on two places: the codacy/codacy-plugins-api and also on qamine/codacy-website (this one is what the Codacy API uses to validate the languages it receives on the endpoint requests). We created this issue to ensure that the lists are synced for now: https://codacy.atlassian.net/browse/CY-2686

docs/generating-coverage-reports.md
pauloribeiro-codacy: Examples would be really nice. I propose to create a new pull request after we merge this one to add some curated links to examples.

docs/alternative-ways-of-running-coverage-reporter.md
pauloribeiro-codacy: I'm not 100% sure about this, but I think I've seen a reference to a customer trusting one of sites more than the other. Could this also be a reason for us to provide two alternatives? :thinking:

docs/alternative-ways-of-running-coverage-reporter.md
pauloribeiro-codacy: Could you provide details about what should be changed? And can this improvement be done on a separate pull request?

docs/adding-coverage-to-your-repository.md
pauloribeiro-codacy: To simplify, I've removed the links to the tools generating the reports.

@mrfyda, your comment suggests that ideally we would provide a combination of languages / testing frameworks / build system. However, it would be tricky to list all the possible combinations that end up generating coverage formats that the Coverage Reporter can parse.

My assumption here is that the file name patterns for the generated reports should be enough to help the users understand if their reports are going to be accepted or not.

mrfyda: I think this will also help: https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r483204395

docs/alternative-ways-of-running-coverage-reporter.md
josemiguelmelo: I think you're right @franciscodua. I also have the idea we changed to bintray because of Github's rate limiting.

@pauloribeiro-codacy I think we should confirm if Github's rate limiting is a problem. If not, we can use that metric to decide which one to use.

pauloribeiro-codacy: I think we are hitting the 60 request per hour API call rate limit, since the script makes an unauthenticated request to the GitHub API to obtain the URL of the binary to download:

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting

Indeed, if customers configure their CI workflows to download the binary using this script, we can easily exceed the limit.

However, it's possible that Bintray also limits requests to their API. Their documentation isn't very conclusive, though:

https://bintray.com/docs/api/#_limits

andreaTP: - Bitbucket is used from the automated script to avoid rate-limiting
 - Github is good for keeping the "Release" page attached to this repository

So far we haven't had much maintenance trouble, so I would keep the two.

pauloribeiro-codacy: Makes sense, and we can always re-evaluate in the future if the need arises. :+1:

pauloribeiro-codacy:
After merging this pull request (whose main focus was on refactoring the information architecture so the users can easily find what they need and follow the instructions) I will follow up the pending tasks or improvements to the documentation identified in this pull request:

- (https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r482852026) Remove list of supported formats from codacy/coverage-parser and redirect to the codacy/codacy-coverage-reporter documentation instead.
- (https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r482853597) Update section on Travis CI (feedback from @mrfyda "Looks like this section could be updated").
- (https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r481299879) Add list of repositories with examples of how to generate coverage (feedback from @mrfyda "Ideally, for each language we would have an example of what it takes to set up coverage and report it to Codacy. See here an example: https://github.com/codecov/example-scala").
- (https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r480306611) Figure out a better solution to list the aliases of the currently supported languages.

franciscodua (approved):

PR #246, opened 2020-08-27
style: Improve error message when there is no API token defined

pauloribeiro-codacy:
Makes it more explicit that:
- A "project token" is a project API token
- An "api token" is an account API token


pauloribeiro-codacy:
I'm closing this pull request and opening a new one (https://github.com/codacy/codacy-coverage-reporter/pull/252) to improve not only this log message but some others as well.

PR #247, opened 2020-09-01
fix: More robust string string splitting

andreaTP:
This should be a more principled approach to string splitting to avoid edge cases.


franciscodua (approved):

franciscodua (approved):

andreaTP:
This was still not compatible with Alpine that was the main reason why I opened it in first place.

PR #248, opened 2020-09-01
fix: Make the orb compatible with Alpine Linux executors

andreaTP:
I manually tested it with ubuntu/alpine and curl/wget matrix ...
Not sure how we can automate tests here ...

### DIFF
--- a/orbs/commands/send_report.yml
+++ b/orbs/commands/send_report.yml
@@ -25,10 +25,10 @@ steps:
         export CODACY_REPORTER_VERSION=<< parameters.tool_version >>
         export CODACY_PROJECT_TOKEN=<< parameters.project-token >>
         # comma separated list of report files
-        report_list=<< parameters.coverage-reports >>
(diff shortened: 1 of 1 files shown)


orbs/commands/send_report.yml
```diff
@@ -41,8 +41,17 @@ steps:
         if << parameters.skip >>; then
             skip_option="--skip"
```
lolgab, line 42: Can you adapt the indentation here to be consistent with your changes?

lolgab (changes requested):

orbs/commands/send_report.yml
andreaTP: done, everything is 4 spaces now.

lolgab (approved):

orbs/commands/send_report.yml
lolgab: Thank you :)

PR #251, opened 2020-09-07
doc: Make improvements to the codacy-coverage-reporter documentation

pauloribeiro-codacy:
This pull request is for working on subsequent improvements to the codacy-coverage-reporter documentation identified in the pull request #244:

- Update section on Travis CI (feedback from @mrfyda "Looks like this section could be updated"). https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r482853597
- Add list of repositories with examples of how to generate coverage (feedback from @mrfyda "Ideally, for each language we would have an example of what it takes to set up coverage and report it to Codacy. See here an example: https://github.com/codecov/example-scala"). https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r481299879
- Figure out a better solution to list the aliases of the currently supported languages. https://github.com/codacy/codacy-coverage-reporter/pull/244#discussion_r480306611

### DIFF
--- a/docs/adding-coverage-to-your-repository.md
+++ b/docs/adding-coverage-to-your-repository.md
@@ -4,7 +4,7 @@ Before setting up Codacy to display code coverage metrics for your repository yo
 
 Codacy supports the following coverage report formats:
 
-| Report formats                | Report file names            |
+| Report format                 | Report file name             |
 | ----------------------------- | ---------------------------- |
 | Clover                        | clover.xml                   |
 | Cobertura                     | cobertura.xml                |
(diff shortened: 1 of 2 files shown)


pauloribeiro-codacy:
@mrfyda, can you provide more details why you think that the section on [how to run the codacy-coverage-reporter on Travis CI](https://github.com/codacy/codacy-coverage-reporter/blob/doc/improve-docs/docs/alternative-ways-of-running-coverage-reporter.md#travis-ci) needs to be updated?

I'm guessing that instead of downloading and running the Java binary, we could use the more automatic script `https://coverage.codacy.com/get.sh `instead. :thinking:

pauloribeiro-codacy:
@mrfyda and @andreaTP, both of you suggested that we provide a few public repositories as examples on how to set up coverage, perhaps for the most popular languages. It would be very nice if we could provide something similar to what Codecov provides:

https://docs.codecov.io/docs/supported-languages

However, this also seems like something that would demand a bit of effort to set up and maintain, and perhaps we can get away with including pointers to 3rd party resources. What do you think?

pauloribeiro-codacy:
@andreaTP and @codacy/toss, we could also discuss here if there is a better solution that we could try out to let the customers know which "programming language aliases" they can use together with the flag `-l`. Right now we're providing a link to [this list in the source code of the codacy-plugins-api](https://github.com/codacy/codacy-plugins-api/blob/master/src/main/scala/com/codacy/plugins/api/languages/Language.scala#L43), however, in the future the list may move to a different line (or even different location altogether) and the link in the documentation will be broken.

franciscodua (approved):

franciscodua (approved):

pauloribeiro-codacy:
I'm going to merge the changes in this pull request because on second thought, the initial scope that I was defining for the pull request is too big and this pull request is already becoming old. Thanks @franciscodua!

PR #253, opened 2020-09-09
doc: Direct users to instructions on how to upload coverage results

pauloribeiro-codacy:
The command provided was incomplete because it missed the `report` instruction. It is better to direct users to the main instructions on how to run codacy-coverage-reporter instead.


franciscodua (approved):

pauloribeiro-codacy:
@franciscodua called my attention to the fact that the command would work without the `report` (it defaults to uploading the coverage data when no command is specified). However, it still makes sense to keep the instructions on how to run the codacy-coverage-reporter in one centralized place.

PR #255, opened 2020-09-21
Allow for user api token

petero-dk:
Use project_token, project_name and username only if defined

### DIFF
--- a/orbs/commands/send_report.yml
+++ b/orbs/commands/send_report.yml
@@ -9,6 +9,14 @@ parameters:
     type: string
     default: ${CODACY_PROJECT_TOKEN}
     description: Specify Codacy's project token
+  project-name:
+    type: string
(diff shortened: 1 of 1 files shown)


machadoit:
Looks great @petero-dk ! However, I think you missed the CODACY_API_TOKEN, right? Let me also test this changes

machadoit:
@petero-dk we are also having to check some details internally regarding the API token usage, since it seems that it is only working for legacy repositories. Probably not going to merge this PR until we receive updates on that end.

machadoit (changes requested):
Marking this PR as needing changes so it is not merged until we have updates from the product side

machadoit:
@petero-dk thank you for your contribution, but we are going to close this PR for now, as we don't have any ETA on when/how we want to tackle this

PR #256, opened 2020-09-22
doc: Remove instructions on how to authenticate using account API Token

pauloribeiro-codacy:
Using the account API Token only works for repositories belonging to legacy organizations and not for synced organizations.

We may add new instructions in the future if we figure out a different solution to submit coverage for multiple repositories without having to use individual project API Tokens for each repository.

See https://codacy.atlassian.net/browse/CY-2862 for more details.

### DIFF
--- a/docs/adding-coverage-to-your-repository.md
+++ b/docs/adding-coverage-to-your-repository.md
@@ -21,23 +21,11 @@ After having coverage reports set up for your repository, you must use Codacy Co
 
 1.  You must set up an API token to allow Codacy Coverage Reporter to authenticate on Codacy.
 
-    -   **If you're setting up coverage for one repository**, obtain the [project API Token](/repositories-configure/integrations/project-api/) from the page **Integrations** in your Codacy repository settings.
+    Obtain the [project API Token](/repositories-configure/integrations/project-api/) from the page **Integrations** in your Codacy repository settings. Then, set the following environment variable to specify your project API Token:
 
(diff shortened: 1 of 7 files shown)


lolgab (approved):

machadoit (dismissed):

rtfpessoa (dismissed):

pauloribeiro-codacy:
I'm thinking that we should only remove the reference to the account API Token from the documentation if we also remove it from the codacy-coverage-reporter itself, because the users will see references to this way of authenticating on the error messages and if they search for more information they won't find it:

https://github.com/codacy/codacy-coverage-reporter/blob/40a8c48a6f7aecd03473151cb570fec2acdd10a8/src/main/scala/com/codacy/rules/ConfigurationRules.scala#L92

machadoit:
I agree with updating the message @pauloribeiro-codacy 👍

pauloribeiro-codacy:
> I agree with updating the message @pauloribeiro-codacy

I will update the error messages and ask for the review of the team. :+1:

pauloribeiro-codacy:
Please check my latest commit (https://github.com/codacy/codacy-coverage-reporter/pull/256/commits/8d84e832205c3072c6158f74aec0c590e8bdb53d) where I remove references to the account API Token from the console output. The intended effect is that we continue to support configurations that already use the account API Token, but we do not provide wrong information for new customers using synced organizations.

franciscodua (approved):

PR #264, opened 2020-10-15
Download command in script

lolgab:
This PR started considering that in Travis CI the `after_success` phase runs after the cache has been saved.
So to cache the binary we need a separate phase that just downloads the binary without sending any coverage.
This also caches the binary based on the version so we're sure the binary gets updated when a new version is out.

### DIFF
--- a/docs/alternative-ways-of-running-coverage-reporter.md
+++ b/docs/alternative-ways-of-running-coverage-reporter.md
@@ -128,12 +128,11 @@ Thanks to the amazing job of [Gavin Mogan](https://github.com/halkeye) you can n
 If you are using Travis CI to report coverage, update your file `.travis.yml` to include the following blocks:
 
 ```yaml
-before_install:
-  - sudo apt-get install jq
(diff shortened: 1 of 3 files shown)


docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -128,12 +128,11 @@ Thanks to the amazing job of [Gavin Mogan](https://github.com/halkeye) you can n
...
 
 ```yaml
-before_install:
-  - sudo apt-get install jq
-  - curl -LSs "$(curl -LSs https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/latest | jq -r '.assets | map({name, browser_download_url} | select(.name | endswith(".jar"))) | .[0].browser_download_url')" -o codacy-coverage-reporter-assembly.jar
+before_script:
+  - bash <(curl -Ls https://coverage.codacy.com/get.sh) download
 
 after_success:
-  - java -jar codacy-coverage-reporter-assembly.jar report -l Java -r build/reports/jacoco/test/jacocoTestReport.xml
```
machadoit, line 136: We don't need specifying the language and path to the report anymore?

machadoit (approved):
LGTM! But there are probably better people to review this than me :D

docs/alternative-ways-of-running-coverage-reporter.md
lolgab: That one was just an example :-)
So I changed the example to be the default one.

franciscodua (approved):

PR #269, opened 2020-11-02
[WIP: Do not merge] feature: Split report by languages CY-3154

lolgab:


machadoit:
Can you describe a bit the behaviour? What it is going to change? Will we be sending multiple reports now? How does Codacy handles that?

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -31,6 +31,10 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
     .asScala
     .map(_.toFile)
 
+  private def flatten[T](v: Either[String, Seq[Either[String, T]]]): Seq[Either[String, T]] = {
```
machadoit, line 34: Why is this a "flatten"? This signature resembles more of a traverse. Is there other file that we can move this method so we can properly document and test?

lolgab: Called it `flatten` because it flattens a `Either[Seq[Either]]` into a `Seq[Either]`.
For what I read, `traverse` usually takes a `Seq` and function from `A => F[B]` returning a transformed `F`.
I created this helper function to avoid duplicating the `.left.map(l => Seq(Left(l))).merge` oneliner.
It seems an overkill to me moving to a separate file and make it public / tested.

Issue #266, opened 2020-10-28
[CY-3145] [Github actions] Can't send coverage due to error


feugy:
### Operating System/Scala version/Java version/Library deps...

Github actions (codacy-coverage-reporter-action's DockerFile)

### Expected Behavior

At the end of the CI job, coverage should be sent to Codacy. 

### Actual Behavior

The coverage file (which exists, and is valid, see attached) is not sent, and the following error is displayed in logs:
```bash
error [CodacyCoverageReporter] Can't guess the language due to invalid path  - (CodacyCoverageReporter.scala:25)

Failed!
```

[coverage file, zipped so it could be attached to that issue](https://github.com/codacy/codacy-coverage-reporter/files/5450644/clover.zip)

[full log attached](https://github.com/codacy/codacy-coverage-reporter/files/5450626/github-action-logs.txt)

### Reproducible Test Case

As my application is open source, you can find failing logs [here](https://github.com/feugy/melodie/runs/1319405928?check_suite_focus=true).

franciscodua:
Hi @feugy 

The problem seems to be an unsupported file extension. In this case, the `.svelte` extension.
For the reporter to send coverage you should exclude these files from the generated coverage.

You can do it by removing the pattern `'**/*.svelte'` from the `collectCoverageFrom` array in `jest.config.js`. (https://github.com/feugy/melodie/blob/master/jest.config.js#L58).

Let us know if it helps. And if this is a good approach to your problem.

feugy:
Hello @franciscodua.
Thanks for this lightning fast response.

I'm sorry but the svelte files are part of the coverage report since day one. 
Svelte is the main technology used, so I can not exclude it.
Besides, Codacy never complained about svelte files in the coverage report until few days ago.

It's only in the latest [PR](https://app.codacy.com/gh/feugy/melodie/pullRequest?prid=6413870) that things broke.
There is no new file extensions in there. Could it be a recent change on Codacy side?

franciscodua:
Hey @feugy 

Ok, I think I know what happened, and it's due to a tool limitation.

This tool assumes the coverage report only has files that use the same language, i.e., all `Javascript`, or all `PHP`. You can check the validation that it's done [here](https://github.com/codacy/codacy-coverage-reporter/blob/master/src/main/scala/com/codacy/rules/ReportRules.scala#L182). It's just checking the first file in the list to check if the tool "knows" the language.

[...]

PR #270, opened 2020-11-04
fix: Improve language guessing by stopping at first supported language CY-3145

lolgab:
Currently the behaviour is to fail guessing language if the first file has an extension we don't support.
This change makes the guesser continue to iterate on file until it finds a file for a language Codacy supports and uses that language file.
This is a temporary workaround until a more complete ( something like https://github.com/codacy/codacy-coverage-reporter/pull/269 ) lands.

### DIFF
--- a/src/main/scala/com/codacy/rules/ReportRules.scala
+++ b/src/main/scala/com/codacy/rules/ReportRules.scala
@@ -179,13 +179,9 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
     languageOpt match {
       case Some(l) => Right(l)
       case None =>
-        report.fileReports.headOption match {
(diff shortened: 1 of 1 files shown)


machadoit:
Can you be more clear on the description? "Not stopping at first language" seems misleading. What your change did was "No longer guesses a language that we don't support", right? You still stop at the first supported language with your change, correct?

lolgab:
Yes, the description is how you say. It iterates over the files and instead of failing it uses the first supported language it finds.
> You still stop at the first supported language with your change, correct?

Correct.
This is far from ideal. But still can offer a temporary workaround for https://github.com/codacy/codacy-coverage-reporter/issues/266 until something better lands.

machadoit (approved):

franciscodua (approved):

Issue #276, opened 2020-12-28
[CY-3399] codacy-coverage-reporter is crashing on macOS Big Sur


hyde1:
- **Operating System**: macOS 11.1

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. Launch the last version of the reporter with `bash <(curl -Ls https://coverage.codacy.com/get.sh) --help`
2. It should display the help

### Actual Behavior

1. Launch the last version of the reporter with `bash <(curl -Ls https://coverage.codacy.com/get.sh) --help`
2. Have the following output

```
     ______          __
    / ____/___  ____/ /___ ________  __
   / /   / __ \/ __  / __ `/ ___/ / / /
  / /___/ /_/ / /_/ / /_/ / /__/ /_/ /
  \____/\____/\__,_/\__,_/\___/\__, /
                              /____/

  Codacy Coverage Reporter

 --> Download the codacy reporter codacy-coverage-reporter-darwin... (latest)
#=#=-  #       #                                                                                                                                                                                                                                               #=O=#     #        #                                                                                                                                                                                                                                           -#O=- #      #          #                                                                                                           ######################################################################################################################################################################################################################################################### 100.0%
######################################################################################################################################################################################################################################################### 100.0%
/dev/fd/11: line 125: 99512 Killed: 9               /Users/hyde/Library/Caches/Codacy/codacy-coverage-reporter --help

 --> Failed!
```
### Reproducible Test Case

Running the last version of the native binary directly also fails ([version 11.4.1](https://dl.bintray.com/codacy/Binaries/11.4.1/codacy-coverage-reporter-darwin)):
```bash
$ ./codacy-coverage-reporter --help                                                                                                                                                                                                                               
zsh: killed     ./codacy-coverage-reporter --help
```

lolgab:
Hi @hyde1,
Thanks for your bug report. Currently, [graalvm native-image doesn't work on Big Sur](https://github.com/oracle/graal/issues/2984).
The issue is solved upstream but a new version with the fix has not been released yet.
As a workaround you could try to [use the java binary](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#manually-downloading-the-java-binary).

lolgab:
This is fixed in the latest version.
Please delete the cache if you want to update (`~/.cache/codacy` on linux and `~/Library/Caches/Codacy` on Mac OS X)

PR #278, opened 2021-01-05
doc: Refactor how to run Codacy Coverage Reporter

pauloribeiro-codacy:
Currently, from the README it's not very clear that there are multiple ways to run the Codacy Coverage Reporter. In particular, users who cannot run the Coverage Reporter on macOS Big Sur have trouble figuring out that they can use the Java binary as a workaround.

See https://github.com/codacy/codacy-coverage-reporter/issues/276 and [CY-3404](https://codacy.atlassian.net/browse/CY-3404) for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ Follow the instructions on how to [add coverage to your repository](https://docs
 
 If necessary, see [alternative ways of running Codacy Coverage Reporter](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/) for other ways of running Codacy Coverage Reporter, such as by installing the binary manually or using a CircleCI Orb or GitHub Action.
 
+**NOTE: If you're using macOS Big Sur** you must [use the Java binary](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#java) to run Codacy Coverage Reporter. For more details, see the issue [#276](https://github.com/codacy/codacy-coverage-reporter/issues/276).
+
 For a complete list of commands and options, run the Codacy Coverage Reporter with the flag `--help`. For example:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
 
 Follow the instructions on how to [add coverage to your repository](https://docs.codacy.com/coverage-reporter/adding-coverage-to-your-repository/).
 
+If necessary, see [alternative ways of running Codacy Coverage Reporter](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/) for other ways of running Codacy Coverage Reporter, such as by installing the binary manually or using a CircleCI Orb or GitHub Action.
+
 For a complete list of commands and options, run the Codacy Coverage Reporter with the flag `--help`. For example:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -11,10 +11,35 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
 
 Follow the instructions on how to [add coverage to your repository](https://docs.codacy.com/coverage-reporter/adding-coverage-to-your-repository/).
 
-For a complete list of commands and options, run:
+For a complete list of commands and options, run the Codacy Coverage Reporter with the flag `--help`. For example:
 
 ```bash
-bash <(curl -Ls https://coverage.codacy.com/get.sh) --help
+$ bash <(curl -Ls https://coverage.codacy.com/get.sh) report --help
+     ______          __
+    / ____/___  ____/ /___ ________  __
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -11,10 +11,39 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
 
 Follow the instructions on how to [add coverage to your repository](https://docs.codacy.com/coverage-reporter/adding-coverage-to-your-repository/).
 
-For a complete list of commands and options, run:
+If necessary, see [alternative ways of running Codacy Coverage Reporter](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/) for other ways of running Codacy Coverage Reporter, such as by installing the binary manually or using a CircleCI Orb or GitHub Action.
```
pauloribeiro-codacy, line 14: After thinking about this, I realized that it's best to keep all the ways in which we can set up and run the Codacy Coverage Reporter on the same page. So, I added this link here to let the users know that they have other ways of running the Coverage Reporter.

README.md
```diff
@@ -11,10 +11,39 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
...
+
+**NOTE: If you're using macOS Big Sur** you must [use the Java binary](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#java) to run Codacy Coverage Reporter. For more details, see the issue [#276](https://github.com/codacy/codacy-coverage-reporter/issues/276).
```
pauloribeiro-codacy, line 16: We should make sure that we remove these warnings when the issue is fixed:

- https://github.com/codacy/codacy-coverage-reporter/pull/278/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R16
- https://github.com/codacy/codacy-coverage-reporter/pull/278/files#diff-8822c8c2af34cf045623870b3ecb526998c4a29a0c859b3cfa1921fb0c11cac6R52
- https://github.com/codacy/codacy-coverage-reporter/pull/278/files#diff-fb26bb713a20eed74104b63774ad41945bc6dae00cd39fa74f392e192248d122R17

pauloribeiro-codacy:
@lolgab, after this pull request is merged we must update the link on the issue https://github.com/codacy/codacy-coverage-reporter/issues/276 to https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#java as the anchor link to the Java section will be different. :arrows_counterclockwise:

lolgab (approved):
Cool improvements 👏👏👏

README.md
```diff
@@ -11,10 +11,39 @@ Multi-language coverage reporter for Codacy https://www.codacy.com
...
+
+For a complete list of commands and options, run the Codacy Coverage Reporter with the flag `--help`. For example:
 
 ```bash
-bash <(curl -Ls https://coverage.codacy.com/get.sh) --help
+$ bash <(curl -Ls https://coverage.codacy.com/get.sh) report --help
+     ______          __
```
jllopes, line 22: Do you think having the output here is valuable? It will easily get outdated

pauloribeiro-codacy: This is a good point @jllopes.

Having the available flags directly on the README of a tool can sometimes give a good feeling of what the tool does and the available configuration options.

Ideally, this would be a more expanded list [like we have for the Codacy Analysis CLI](https://github.com/codacy/codacy-analysis-cli#commands-and-configuration), for example, but I thought that as a first approach the direct output from the `--help` switch could be enough.

However, if the available options tend to be frequently updated, then it's well possible that it's not worth paying the maintenance cost of also having to update the README. What do you think?

machadoit: I do tend to agree with @jllopes point, but in this case in particular I think the Pros outweigh the Cons on showing the information on the repo. So I would suggest to maintain the suggestion of adding this information on the repo, but I can behind both approaches.

README.md
```diff
@@ -24,10 +53,10 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) --help
 ### Among Codacy’s features:
 
 - Identify new Static Analysis issues
-- Commit and Pull Request Analysis with GitHub, BitBucket/Stash, GitLab (and also direct git repositories)
+- Commit and Pull Request Analysis with GitHub, GitLab, and Bitbucket
 - Auto-comments on Commits and Pull Requests
-- Integrations with Slack, HipChat, Jira, YouTrack
```
machadoit, line 29: 👍

machadoit (approved):

README.md
pauloribeiro-codacy: We can always decide on a different approach if, later on, we find out that this adds an extra maintenance chore. :+1:

jllopes: Yes, let's keep it for now and we'll see

jllopes (approved):

PR #282, opened 2021-01-15
fix: Strip trailing "/" from CODACY_API_BASE_URL environment variable

pauloribeiro-codacy:
On https://github.com/codacy/docs/issues/404 a user noticed that the upload fails if the defined `CODACY_API_BASE_URL` environment variable includes a trailing slash character (`/`).

This is a "quick and dirty" suggestion to make the Codacy Coverage Reporter work both in this scenario and also when the base URL is specified with the flag `--codacy-api-base-url`.

### DIFF
--- a/src/main/scala/com/codacy/rules/ConfigurationRules.scala
+++ b/src/main/scala/com/codacy/rules/ConfigurationRules.scala
@@ -77,7 +77,7 @@ class ConfigurationRules(cmdConfig: CommandConfiguration, envVars: Map[String, S
       commitUUID <- validateCommitUUIDConfig(baseConfig)
       baseConf = BaseConfig(
         authConfig,
-        baseConfig.codacyApiBaseUrl.getOrElse(getApiBaseUrl),
+        baseConfig.codacyApiBaseUrl.getOrElse(getApiBaseUrl).replaceAll("/$", ""),
         commitUUID,
         baseConfig.debugValue
       )


pauloribeiro-codacy:
For some reason, [this step of the CI/CD pipeline](https://app.circleci.com/pipelines/github/codacy/codacy-coverage-reporter/795/workflows/b633d1fb-212f-4a3c-a843-7580d1cdb8de/jobs/6617) failed, but it's also failing on other branches. :thinking:

lolgab (changes requested):
This problem happens because we're concatenating paths as strings in codacy-api-scala instead of using a proper path interpolation.
This workaround is not needed if we solve the problem upstream.

The string interpolation happens [here](https://github.com/codacy/codacy-api-scala/blob/3c4a95dccd75f14371171052d5c0f4215998c1b3/src/main/scala/com/codacy/api/client/CodacyClient.scala#L28)

pauloribeiro-codacy:
> This problem happens because we're concatenating paths as strings in codacy-api-scala instead of using a proper path interpolation.
> This workaround is not needed if we solve the problem upstream.
> 
> The string interpolation happens [here](https://github.com/codacy/codacy-api-scala/blob/3c4a95dccd75f14371171052d5c0f4215998c1b3/src/main/scala/com/codacy/api/client/CodacyClient.scala#L28)

Good suggestion @lolgab. I'm closing this pull request then, as I've opened https://github.com/codacy/codacy-api-scala/pull/42 where we can improve the URL building.

PR #287, opened 2021-01-26
bump: Update Graalvm to 21.0.0

lolgab:


pedrocodacy (approved):

build.sbt
```diff
@@ -61,38 +61,25 @@ cancelable in Global := true
 javacOptions ++= Seq("-source", "11", "-target", "11")
 
 enablePlugins(GraalVMNativeImagePlugin)
-graalVMNativeImageGraalVersion := Some("20.0.0-java11")
+GraalVMNativeImage / containerBuildImage := Some("graavlm:latest")
```
machadoit, line 64: latest? Are we sure? Don't we want to fix a version?

build.sbt
```diff
@@ -61,38 +61,25 @@ cancelable in Global := true
...
 
 graalVMNativeImageOptions := Seq(
   "--verbose",
   "--no-server",
   "--enable-http",
   "--enable-https",
-  "--enable-url-protocols=http,https,file,jar",
+  "--enable-url-protocols=http,https,jar",
```
machadoit, line 71: does it mean anything? to remove this?

machadoit (approved):

build.sbt
lolgab: This image we are creating ourselves in the CI with the Dockerfile included.
We need to set a different image because the sbt plugin uses a fixed image name "oracle/graavm-ce" image that doesn't exist anymore.

build.sbt
lolgab: If you try to generate the image with graalvm 21.0.0 you get a deprecation warning saying that this is the new default and you don't need to specify it anymore

PR #291, opened 2021-02-01
fix: Fix CircleCI orb when version is not defined

lolgab:
Now that cache is versioned the latest version is chosen only when `CODACY_REPORTER_VERSION` is empty \([here](https://github.com/codacy/codacy-coverage-reporter/blob/master/get.sh#L106)\).

The integration test that failed is flaky.

### DIFF
--- a/orbs/commands/send_report.yml
+++ b/orbs/commands/send_report.yml
@@ -3,7 +3,7 @@ description: "Download Codacy's coverage reporter and run it"
 parameters:
   tool_version:
     type: string
-    default: latest
+    default: ''
     description: "Specify Codacy's coverage reporter tool version"
   project-token:
     type: string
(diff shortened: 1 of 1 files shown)


franciscodua (approved):

orbs/commands/send_report.yml
```diff
@@ -3,7 +3,7 @@ description: "Download Codacy's coverage reporter and run it"
 parameters:
   tool_version:
     type: string
-    default: latest
```
machadoit, line 6: The '' working makes sense, as if there is no default passed, the tool should fall back to latest. However, I think that we should make changes on the direction of supporting latest, don't you think?

Because, if people are passing explicitly latest at the moment, it will fail for them, correct?

lolgab: > Because, if people are passing explicitly latest at the moment, it will fail for them, correct?

Correct. However, latest is a pretty arbitrary name to say "contact the github api to find the tags and take the last one".
I would just not define and avoid special treatments, I doubt someone manually set that to "latest".
What do you think?

jllopes (approved):

orbs/commands/send_report.yml
lolgab: Anyway I'm merging this as is to unblock all the users that use the default setting.
I can change this in a follow up PR.

PR #301, opened 2021-03-05
feature: Add workaround for MalformedInputException DOCS-217

pauloribeiro-codacy:
After the investigation in [CY-3801](https://codacy.atlassian.net/browse/CY-3801) we came to the conclusion that the recommended workaround for [this issue](https://community.codacy.com/t/issue-getting-malformedinputexception-input-length-1-when-uploading-lcov-info-from-github-actions/181) would be to make sure that the coverage report is encoded using UTF-8.

### DIFF
--- a/docs/troubleshooting-common-issues.md
+++ b/docs/troubleshooting-common-issues.md
@@ -69,6 +69,12 @@ dotCover.exe cover ... /Filters=-:xunit*
 
 By default, dotCover includes xUnit files in the coverage analysis and this results in larger coverage reports. This filter helps ensure that the resulting coverage data does not exceed the size limit accepted by the Codacy API when uploading the results.
 
+## MalformedInputException while parsing report
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting-common-issues.md
```diff
@@ -69,6 +69,12 @@ dotCover.exe cover ... /Filters=-:xunit*
 
 By default, dotCover includes xUnit files in the coverage analysis and this results in larger coverage reports. This filter helps ensure that the resulting coverage data does not exceed the size limit accepted by the Codacy API when uploading the results.
 
+## MalformedInputException while parsing report
+
+If you get a `java.nio.charset.MalformedInputException` when running the Codacy Coverage Reporter it means that the coverage report includes an unsupported character, perhaps on one of your source code file names.
+
+For maximum compatibility of your coverage reports with the Codacy Coverage Reporter, make sure that your coverage reports use UTF-8 encoding or remove any special characters from the reports.
```
pauloribeiro-codacy, line 76: > [...] or remove any special characters from the reports

I think it could still be a good idea to try to avoid special characters in source code file names, not just for compatibility with the Codacy Coverage Reporter. :thinking:

PR #304, opened 2021-03-18
doc: Add simplified way of uploading multiple coverage reports

pauloribeiro-codacy:
It's possible to upload multiple test coverage reports using a single command. This also makes it easier to find and specify multiple report files dynamically.

:construction: **Should we keep the information in red or is it just adding unnecessary complexity in the documentation?** Plase see the comments below.

image

See [this Slack thread](https://codacy.slack.com/archives/C8X9SS1H7/p1615805679001700) for more details.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -154,7 +154,7 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
     -l Java -r report1.xml -r report2.xml -r report3.xml
 ```
 
-You can also specify the reports to upload dynamically using the command `find`. For example:
+You can also upload all your reports dynamically using the command `find`. For example:
 
 ```bash
 bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
--- a/docs/index.md
+++ b/docs/index.md
@@ -145,28 +145,34 @@ See the sections below for more advanced functionality, or [check the troublesho
 
 ## Uploading multiple coverage reports for the same language {: id="multiple-reports"}
 
-If your test suite is split on different modules or runs in parallel, you will need to upload multiple coverage reports for the same language:
+If your test suite is split on different modules or runs in parallel, you will need to upload multiple coverage reports for the same language.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -145,28 +145,34 @@ See the sections below for more advanced functionality, or [check the troublesho
...
-For example:
+You can also upload all your reports dynamically using the command `find`. For example:
 
 ```bash
 bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
-    --partial -l Java -r report1.xml
-bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
-    --partial -l Java -r report2.xml
-bash <(curl -Ls https://coverage.codacy.com/get.sh) final
+    -l Java $(find **/jacoco*.xml -printf '-r %p ')
```
pauloribeiro-codacy, line 161: @heliocodacy and @IGFCoimbra, from your interactions with customers, would you say that this use case is frequent and useful enough to be mentioned on the documentation - with the risk of adding a bit more complexity for users who don't need this?

docs/index.md
```diff
@@ -145,28 +145,34 @@ See the sections below for more advanced functionality, or [check the troublesho
...
+
+    ```bash
+    bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
+        --partial -l Java -r report1.xml
+    bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
+        --partial -l Java -r report2.xml
+    bash <(curl -Ls https://coverage.codacy.com/get.sh) final
+    ```
+
+    If you're sending reports for a language with the flag `--partial`, you should use the flag in all reports for that language to ensure the correct calculation of the coverage.
```
pauloribeiro-codacy, line 175: Are there scenarios where this way of uploading the coverage report data is more convenient? Do we need to keep these instructions or perhaps we can simplify and include only the solution of uploading all reports with a single command?

IGFCoimbra (approved):

PR #306, opened 2021-03-24
doc: Simplify instructions on uploading multiple coverage reports DOCS-228

prcr:
After adding instructions on uploading multiple coverage reports using a single command (#304), I propose that we simplify the documentation by removing the instructions for the `--partial` + `final` method, **as long as there isn't a good reason** why customers would need or prefer to upload coverage reports using this method.

Here's a preview of the proposed solution:

image

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -159,7 +159,7 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
 ```
 
 !!! tip
-    You can also upload all your reports dynamically using the command `find`. For example:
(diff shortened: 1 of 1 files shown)


mrfyda (dismissed):

mrfyda (approved):

mrfyda (approved):

prcr:
I'm closing this pull request because in the meantime I've learned that uploading partial coverage reports could be useful for example in a [parallel testing scenario](https://bitbar.com/blog/parallel-testing-what-it-is-and-why-you-should-adopt-it/), where partial coverage reports are uploaded on each test run.

PR #322, opened 2021-05-25
fix: Fix commit uuid for pull_request_target GH event CY-4354

lolgab:
Fix for pull_request_review and pull_request_review_comment too

Tested the four events separately on a dummy repository through SSH with [Debugging with tmate](https://github.com/marketplace/actions/debugging-with-tmate)

### DIFF
--- a/src/main/scala/com/codacy/rules/commituuid/providers/GitHubActionProvider.scala
+++ b/src/main/scala/com/codacy/rules/commituuid/providers/GitHubActionProvider.scala
@@ -17,11 +17,12 @@ object GitHubActionProvider extends CommitUUIDProvider with LazyLogger {
   }
 
   override def getValidCommitUUID(environment: Map[String, String]): Either[String, CommitUUID] = {
-    // if the event is a pull request the GITHUB_SHA will have a different commit UUID (the id of a merge commit)
+    // If the event is a pull request event the GITHUB_SHA will have a different commit UUID (the id of a merge commit).
+    // It happens with all the pull_request* events: pull_request, pull_request_review, pull_request_review_comment and pull_request_target.
     // https://help.github.com/en/actions/reference/events-that-trigger-workflows
     // for this reason, we need to fetch it from the event details that GitHub provides
     // equivalent to doing ${{github.event.pull_request.head.sha}} in a GitHub action workflow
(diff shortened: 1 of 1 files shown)


prcr (commented):
I understand how this solution works :+1: but conceptually, aren't we circumventing or even going against the best practices explained in this article? And can we run the risk of someone marking the Coverage Reporter as "insecure code"?

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Did we already understand why the user is using the particular event `pull_request_target` and if it's really necessary to use it instead of the recommended event `pull_request`?

Just some food for thought so we consider all options before committing to a solution. :+1:

src/main/scala/com/codacy/rules/commituuid/providers/GitHubActionProvider.scala
```diff
@@ -17,11 +17,12 @@ object GitHubActionProvider extends CommitUUIDProvider with LazyLogger {
...
 
   override def getValidCommitUUID(environment: Map[String, String]): Either[String, CommitUUID] = {
-    // if the event is a pull request the GITHUB_SHA will have a different commit UUID (the id of a merge commit)
+    // If the event is a pull request event the GITHUB_SHA will have a different commit UUID (the id of a merge commit).
+    // It happens with all the pull_request* events: pull_request, pull_request_review, pull_request_review_comment and pull_request_target.
     // https://help.github.com/en/actions/reference/events-that-trigger-workflows
     // for this reason, we need to fetch it from the event details that GitHub provides
     // equivalent to doing ${{github.event.pull_request.head.sha}} in a GitHub action workflow
-    if (environment.get("GITHUB_EVENT_NAME").contains("pull_request")) {
+    if (environment.get("GITHUB_EVENT_NAME").exists(_.startsWith("pull_request"))) {
```
machadoit, line 25: Wow, interesting.. Can you fix it at https://github.com/codacy/codacy-analysis-cli-action/blob/master/action.yml#L92 for the CLI as well? Since I assume it has the same problem?

machadoit (approved):

machadoit:
I think that what @prcr saying makes a lot of sense, it seems that we can actually open problems with this, right?

lolgab:
Declined since we don't want to support this use case, for the security problems it can arise.

PR #335, opened 2021-08-12
Check checksum when downloading binaries CY-4457

machadoit:
We already had a CODACY_REPORTER_VERSION and now are adding a CODACY_REPORTER_SKIP_CHECKSUM, any place where you guys think we could mention it? @prcr 

This will check for the checksum of the binaries >= 12.4.0 and only if CODACY_REPORTER_SKIP_CHECKSUM is not set to `true`

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -65,17 +65,42 @@ exit_trap() {
 trap exit_trap EXIT
 trap 'fatal Interrupted' INT
 
-download() {
+download_stdout() {
     local url="$1"
-    local output="${2:--}"
 
     if command -v curl > /dev/null 2>&1; then
-        curl -# -LS "$url" -o "$output"
+        curl -# -LS "$url" -o-
     elif command -v wget > /dev/null 2>&1; then
-        wget "$url" -O "$output"
+        wget "$url" -O-
     else
         fatal "Could not find curl or wget, please install one."
     fi
+}
+
+download_file() {
+    local url="$1"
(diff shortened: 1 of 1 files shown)


lucassklp (approved):

get.sh
```diff
@@ -65,19 +65,63 @@ exit_trap() {
...
+    local output_filename="$4"
+    local checksum_url="$5"
+
+    pushd "$output_folder"
+
+    download_file "$url"
+    checksum "$file_name" "$checksum_url"
+    mv "$file_name" "$output_filename"
+
+    popd
```
tobiasweibel, line 122: @machadoit would it be possible to replace `pushd` and `popd` again since they are an enhancement of POSIX standard in the bourne shell and don't work e.g. on a pure alpine linux.

machadoit: I guess that is a fair point, do you have any suggestion here, or something like this would make sense for you?

```
original_folder="$(pwd)"
cd "$output_folder"

##Do the download and checksum

cd "$original_folder"
```

tobiasweibel: I'm sometimes using `cd path` and later `cd -`. But I'm not so sure whether that's working for all shells.

machadoit: I will be proposing without the `cd -` so it does not need to depend on anything that may happen on the `##Do the download and checksum` now and on the future.

PR #337, opened 2021-08-13
doc: Adds documentation on how to check the checksum of the binaries

machadoit:


docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -85,6 +90,25 @@ You can use the scripts below to automatically check for the latest version of t
     java -jar codacy-coverage-reporter-assembly.jar report
     ```
 
+### \[Optional\] Checksums
+
+For any approach where you download the binary you can manually check the checksum for it.
```
jllopes, line 95:
```suggestion
When downloading the binary you can manually check the checksum for it.
```

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -29,6 +29,11 @@ To use a specific version of the Codacy Coverage Reporter, set the following env
 export CODACY_REPORTER_VERSION=<version>
 ```
 
+> Note that for versions over `13.0.0` the script will check automatically the checksum of the downloaded binary.
+To override this behavior and skip the checksum of the binary you can:
+
+> `export CODACY_REPORTER_SKIP_CHECKSUM=true`
```
prcr, line 35: Simplify text and use an admonition card:

```suggestion
!!! note
    Starting on version `13.0.0` the script automatically validates the checksum of the downloaded binary. To skip the checksum validation, define the following environment variable:

    ```bash
    export CODACY_REPORTER_SKIP_CHECKSUM=true
    ```
```

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -85,6 +90,25 @@ You can use the scripts below to automatically check for the latest version of t
...
+```
+
+For the cheksum check, you can use any tool of your choice, as long as it relies on `SHA512`.
+As an example you can:
+
+```bash
+sha512sum --check codacy-coverage-reporter-linux.SHA512SUM
+```
+
+> Note that this concrete command is looking for the file named `codacy-coverage-reporter-linux` on that
```
jllopes, line 109: I would replace `on that same folder` with `on the current directory`

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -85,6 +90,25 @@ You can use the scripts below to automatically check for the latest version of t
     java -jar codacy-coverage-reporter-assembly.jar report
     ```
 
+### \[Optional\] Checksums
```
prcr, line 93:
```suggestion
### \[Optional\] Validating the checksum of the binaries
```

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -85,6 +90,25 @@ You can use the scripts below to automatically check for the latest version of t
...
+
+```bash
+curl -Ls -O https://github.com/codacy/codacy-coverage-reporter/releases/download/13.0.0/codacy-coverage-reporter-linux.SHA512SUM
+```
+
+For the cheksum check, you can use any tool of your choice, as long as it relies on `SHA512`.
+As an example you can:
+
+```bash
+sha512sum --check codacy-coverage-reporter-linux.SHA512SUM
```
prcr, line 106: This can be made more compact, and the example can include downloading and validating the checksum all at once (this is what customers will probably want to do).

This also fixes the URL for the version 13.3.0. :slightly_smiling_face: 

```suggestion
You can use the checksums [available for each release](https://github.com/codacy/codacy-coverage-reporter/releases) to validate the binaries that you download manually. You can use any tool of your choice to validate the checksum, as long as it uses the `SHA512` algorithm.

For example, run the commands below to download and validate the checksum for the 13.3.0 Linux binary. Note that the command `sha512sum` expects to find the binary on the same directory and with the original name `codacy-coverage-reporter-linux`.

```bash
curl -Ls -O https://github.com/codacy/codacy-coverage-reporter/releases/download/13.3.0/codacy-coverage-reporter-linux.SHA512SUM
sha512sum --check codacy-coverage-reporter-linux.SHA512SUM
```

docs/alternative-ways-of-running-coverage-reporter.md
```diff
@@ -85,6 +90,25 @@ You can use the scripts below to automatically check for the latest version of t
...
+```
+
+> Note that this concrete command is looking for the file named `codacy-coverage-reporter-linux` on that
+same folder. You need to make sure that the downloaded binary has the original name during this step.
+
```
prcr, line 111: This part can be removed if it's mentioned directly in the example above.

```suggestion
```

prcr (commented):
I made a few suggestions to try to make the information more compact and straightforward. :+1:

prcr:
@jllopes, even though I was reviewing the pull request at the same time as you, I think my suggestions also address your feedback. :smile:

docs/alternative-ways-of-running-coverage-reporter.md
prcr: Also, maybe we don't need to mention this is optional? To me at least it has always been self-explanatory that even though it's a good idea to validate the downloaded binaries, one may choose not to do it (and run the risks :sweat_smile: ).

So this could become:

```suggestion
### Validating the checksum of the binaries
```

docs/alternative-ways-of-running-coverage-reporter.md
machadoit: Oops, actually the 13.3.0 was the typo, it was actually 13.0.0, going to update it

prcr (approved):

PR #349, opened 2021-09-02
doc: Include workaround for checksum issues in troubleshooting CY-4874

prcr:
Following the [feedback](https://codacy.atlassian.net/browse/CY-4874?focusedCommentId=47679) from @dan-codacy and other reports from users who have had issues with the checksum calculation, we should make the workaround for disabling the checksum calculation more visible, directly in the existing [troubleshooting page](https://docs.codacy.com/coverage-reporter/troubleshooting-common-issues/).

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -105,7 +105,7 @@ checksum() {
     elif command -v shasum > /dev/null 2>&1; then
         sha_check_command="shasum -a 512"
     else
-        fatal "Error: no method of validating checksum, please install 'sha512sum' or 'shasum'. You can skip this check by setting CODACY_REPORTER_SKIP_CHECKSUM=true"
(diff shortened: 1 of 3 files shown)


get.sh
```diff
@@ -105,7 +105,7 @@ checksum() {
     elif command -v shasum > /dev/null 2>&1; then
         sha_check_command="shasum -a 512"
     else
-        fatal "Error: no method of validating checksum, please install 'sha512sum' or 'shasum'. You can skip this check by setting CODACY_REPORTER_SKIP_CHECKSUM=true"
+        fatal "Error: can't validate checksum, please see https://docs.codacy.com/coverage-reporter/troubleshooting-common-issues/#checksum"
```
prcr, line 108: I propose that we include a direct link to the troubleshooting section because like this we have the freedom to include more details/information to help the users.

machadoit (approved):
Very good, thx!

PR #350, opened 2021-09-20
fix regex for lcov files

aochsner:
codacy never seems to find my lcov reports.  I have to always pass them in directly via `-r` command.  

I'm not a Scala expert but it seems the regex for lcov doesn't like nested capture groups.  

As a result, had to fix test lcov to ensure they aren't found when running codacyCoverage tests


lolgab:
Thank you very much for your contribution @aochsner!
I refactored the code a bit so we don't need to rename the lcov test files to make the test pass.
Other than that your fix just went to master.

Issue #234, opened 2020-07-08
[DOCS-222] Add indication: File paths within the coverage report should match what's in Codacy


paulacodacy:
One frequent issue with coverage related tickets is the mismatch between the file path declared within the coverage report and the file path registered on the Codacy UI > Files page. Example: User has the coverage report with "github.com/<Org_Name>/<Project_name>/cmd/AAA/BBB.go", but on the repository's UI > Files page, the same file follows only the package directory "cmd/AAA/BBB.go". 

This being said, we could consider include an indication in the `codacy-coverage-reporter`'s `README.md` or in the logs to pay attention to this part.

PR #351, opened 2021-09-23
feature: Troubleshooting for coverage missing in Codacy UI DOCS-222

prcr:
Adds a small troubleshooting section for the scenario where the Codacy Coverage Reporter successfully uploads the coverage data but it doesn't become visible on the Codacy UI. The cause can either be a mismatch in the file paths included in the coverage reports or a wrong commit UUID being reported.

Fixes https://github.com/codacy/codacy-coverage-reporter/issues/234.


machadoit (approved):
👍

madalenamelo (approved):

PR #352, opened 2021-09-24
fix: Fix regex for lcov filenames

lolgab:


src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -18,11 +18,12 @@ import scala.collection.JavaConverters._
 class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
 
   private val rootProjectDir = new File(System.getProperty("user.dir"))
-  private val rootProjectDirIterator = Files
+  private val rootProjectFiles = Files
     .walk(rootProjectDir.toPath)
     .iterator()
     .asScala
     .map(_.toFile)
+    .toList
```
machadoit, line 26: should we use view or something here, or how it works the iterator as Scala?

lolgab: `.asScala` converts the java iterator to a Scala iterator which is semantically equivalent to a view in this scope.

src/test/scala/com/codacy/rules/ReportRulesSpec.scala
```diff
@@ -120,35 +130,44 @@ class ReportRulesSpec extends WordSpec with Matchers with PrivateMethodTester wi
...
-      val fileIterator = Iterator(new File("index.xml"), new File("coverage-xml", "index.xml"))
-      val reportEither = components.reportRules.guessReportFiles(List.empty, fileIterator)
+      val projectFiles = List(new File("index.xml"), new File("coverage-xml", "index.xml"))
+      val reportEither = components.reportRules.guessReportFiles(List.empty, projectFiles)
 
       reportEither should be('right)
       reportEither.right.value should be(List(new File("coverage-xml", "index.xml")))
     }
+
+    "find an lcov report" in {
```
machadoit, line 163: Would this test fail without the fix? Meaning, will this prevent a regression? What was failing before that now works?

lolgab: Now we are better at finding lcov files thanks to this fix.

machadoit: What I mean is, if you revert the fix. Will this test fail? It would be cool to have a test that prevented regressions

lolgab: > Would this test fail without the fix?
Yes, it fails because the old regex was not correct.
image

The important fix is the one you see in the first commit, here: https://github.com/codacy/codacy-coverage-reporter/pull/352/files#diff-90073b9c1b8eacea1793b1c07107761d5c1868c40c7b12ce47d34826bd4c4bdcR211

machadoit: As a future reference, the problem is that the old regex had multiple capture groups, and was not escaping the `.`

machadoit (approved):

PR #396, opened 2022-06-28
feature: prevent concurrent ci workflows runs on the same branch

pedrocodacy:
Quick effort to prevent problems publishing the reporter due to concurrently running workflows. This does a similar behaviour from the implementation in the QA suites:

- introduce a new step in the beginning of the workflow to actively wait for a slot to run (through python script)
- poll circleci API for existing pipelines for the current branch
- if any exist check if their workflows are in running state
- if any, wait and try again in some minutes

This can be followed up by abstracting this to `codacy-orbs`

### DIFF
--- a/.circleci/canWorkflowStart.py
+++ b/.circleci/canWorkflowStart.py
@@ -1,0 +1,154 @@
+#!/usr/bin/python3
+
+# This script checks if any circleci job is currently running in a target branch.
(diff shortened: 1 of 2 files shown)


h314to (approved):
Nice 👍 I'll be adding a step to do this to the orb.

PR #397, opened 2022-06-30
doc: Add extra context during coverage setup DOCS-386

prcr:
Adds extra context and information that is important to clarify how Codacy works with coverage and to troubleshoot scenarios where Codacy may not display the coverage metrics even though the users have completed the coverage setup instructions.

Fixes https://github.com/codacy/docs/issues/1268.

### :eyes: Live preview
https://feature-update-codacy-coverage-repo--docs-codacy.netlify.app/coverage-reporter/


docs/index.md
```diff
@@ -269,3 +275,32 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
...
+
+Codacy displays the code coverage in each branch, as well as the evolution of code coverage between commits and the code coverage variation introduced by pull requests. Because of this, to ensure that all code coverage metrics are available on Codacy, you must have successfully uploaded coverage data and analyzed:
+
+-   The last two commits in each branch
+-   The common ancestor commits between pull request branches and their target branches
+
+The example below shows that after pushing a commit that correctly sets up coverage on the main branch:
+
+-   Codacy will be able to report the coverage metrics for all subsequent commits and pull requests relative to the main branch
+-   However, Codacy won't be able to report coverage for commits and pull requests that are relative to older branches where the coverage setup wasn't performed yet
```
lventura-codacy, line 289: Personally I'd remove the however, and keep it as "Codacy won't be able to report coverage for commits and pull requests that are relative to older branches where the coverage setup wasn't performed yet".
Note that I'm assuming each bullet completes the sentence before the bullet points, is this correct?

docs/index.md
```diff
@@ -269,3 +275,32 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
...
+To validate that the coverage setup is complete:
+
+1.  Wait until there are **at least two commits** sending coverage data that are analyzed by Codacy.
+
+    !!! important
+        Codacy only takes the uploaded coverage data into account after successfully analyzing each commit.
+
+        Make sure that you [invite or ask your team members to join your organization on Codacy](../organizations/managing-people/#adding-people) so that Codacy analyzes their commits on private repositories.
+
+1.  Check that Codacy displays the coverage information for the latest commits and pull requests, either as a positive, negative, or no variation (represented by `=`) of the coverage percentage:
```
lventura-codacy, line 302: Should we mention that diff coverage can be a `∅` sign?

lventura-codacy (commented):
Left some minor comments, but otherwise LGTM!

docs/index.md
prcr: You're right, each item in the list should complete the introductory sentence. :+1: 

I was trying to emphasize the difference between the two statements, so besides removing the "However" I also simplified the beginning of each list item and I believe that now the opposition between the statements is quite clear.

docs/index.md
```diff
@@ -269,3 +275,32 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
...
+The example below shows that after pushing a commit that correctly sets up coverage on the main branch:
+
+-   Codacy will be able to report the coverage metrics for all subsequent commits and pull requests relative to the main branch
+-   However, Codacy won't be able to report coverage for commits and pull requests that are relative to older branches where the coverage setup wasn't performed yet
+
+![Setting up coverage on the main branch](images/coverage-validate.png)
+
+To validate that the coverage setup is complete:
+
+1.  Wait until there are **at least two commits** sending coverage data that are analyzed by Codacy.
```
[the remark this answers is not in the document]
prcr: Good point! I edited the sentence to (hopefully) make everything more explicit. :+1:

docs/index.md
prcr: Good catch!

Here, I actually decided to simplify and mention only the `-` symbol in the scenario where the coverage isn't displayed, since the other values depend if the metric is the diff coverage or the coverage variation. Besides that, the diff coverage is only available for pull requests.

I also realized that it's best not to enumerate the possible values for coverage in multiple documentation pages, because it's easy for the information to become outdated. It's better if we stick to a single "source of truth".

IGFCoimbra (approved):

madalenamelo (dismissed):

PR #398, opened 2022-07-06
Use orb step to wait for workflows

h314to:
This uses the new step in the orb to wait for other running workflows, thus dispensing the need to have a script in the project doing this.

Additionally, by using the latest version of the codacy-orb it also uses the latest stable version of xcode supported by circleci, thus avoiding [an issue caused by the deprecation of the xcode version currently in use](https://discuss.circleci.com/t/xcode-image-deprecation/44294).


lolgab:
Can this be a step inside `checkout_and_version` so we can avoid the overhead of another job?

h314to:
> Can this be a step inside `checkout_and_version` so we can avoid the overhead of another job?

I think having a separate step is nice because it gives you the flexibility to check for running workflows at different times of the build. For instance, say your build/test steps can run independently of other ci jobs. These are often the bulk of the work in CI and where developers might get some feedback of stuff that needs changing/fixing, so it is nice if you can do this irrespective of any other worflows running. Thus, you can postpone the check for other worflows running later, say when publishing artifacts or deploying, which is really where things can brake if something else is running at the same time.

I'm placing the wait at the beginning of the workflow because that is where the previous step that did this was, but I think it would be beneficial for your workflow if you can move it further down the workflow.

Moreover, you'll be using a wait step so you're bound to be wasting some time anyways. I think the time optimisation should instead be done by making sure that you can run workflows concurrently without issues.

That being said, we can always add the wait functionality to checkout and version and refactor this later on, and there would be some time saved if you really want/need to keep this step in the beginning of the workflow, but I think we need to have and individual step nonetheless.

lolgab (approved):
LGTM! 👍

PR #400, opened 2022-07-06
Add retry process when sending coverage CY-6106

stefanvacareanu7:


stefanvacareanu7:
I still need to test it properly, create new tests and see if this is the method I should go with.

README.md
```diff
@@ -37,6 +37,8 @@ Usage: codacy-coverage-reporter report
   --commit-uuid  <your commitUUID>
   --http-timeout  <Sets a specified read timeout value, in milliseconds, to be used when interacting with Codacy API>
   --skip | -s  <skip if token isn't defined>
+  --sleep-time <Sets a specified time, in milliseconds, to ne used when waiting between retries>
```
lventura-codacy, line 40:
```suggestion
  --sleep-time <Sets a specified time, in milliseconds, to be used when waiting between retries>
```

src/it/scala/com/codacy/CodacyCoverageReporterSpec.scala
```diff
@@ -33,7 +35,9 @@ class CodacyCoverageReporterSpec extends WordSpec with Matchers with EitherValue
         username,
         projectName,
         apiBaseUrl,
-        commitUuid
+        commitUuid,
+        sleepTime, // WHY requiring Int if in the BaseCommandConfig I declared as Option[Int]
```
lventura-codacy, line 39: The code is compiling with the optionals, right? Is it just IntelliJ acting up?

src/main/scala/com/codacy/configuration/parser/ConfigurationParser.scala
```diff
@@ -88,9 +88,15 @@ case class BaseCommandConfig(
...
     commitUUID: Option[String],
     @ValueDescription(
-      "Sets a specified read timeout value, in milliseconds, to be used when interacting with Codacy API"
+      "Sets a specified read timeout value, in milliseconds, to be used when interacting with Codacy API. By default, the value is 10 seconds"
     )
     httpTimeout: Int = 10000,
+    @ValueDescription(
+      "Sets a specified time, in milliseconds, to be used when waiting between retries. By default, the value is 10 seconds"
+    )
+    sleepTime: Option[Int] = Some(10000),
```
machadoit, line 97: Optionals with default like this are strange to me. Because, is there a way to 'turn off this feature'? I think that either they don't have defaults and you need to pass the things as needed, or just like we did for `httpTimeout` we can just have `sleepTime: Int = 10000` and the same for `numRetries`. What do you think?

machadoit:
SGTM! Nice stuff!

machadoit (approved):

PR #404, opened 2022-08-01
Update README.md

stefanvacareanu7:


stefanvacareanu7:
@prcr The ' from "isn't" together with ```bash is causing that after that ' to have no colors for the closing brackets. This is a very minor change, so we can ignore it if it doesn't make sense

lolgab:
As an alternative you can remove the bash syntax highlighting

stefanvacareanu7:
Sounds good to me as well. Let's do it

prcr (approved):
:shipit:

PR #406, opened 2022-08-03
doc: Refactor coverage documentation

prcr:
The [main page with instructions for setting up coverage](https://docs.codacy.com/coverage-reporter/) is getting too big, making it hard to follow through the complete sequence of instructions.

This pull request:

- Moves the instructions for more advanced scenarios to a dedicated page
- Tweaks a few references and admonitions so that they appear in the sequence that makes the most sense while users are configuring their setup

These improvements will give way to more detailed troubleshooting instructions for the most common coverage setup failure scenarios (see [CY-6263](https://codacy.atlassian.net/browse/CY-6263)).

PR #407, opened 2022-08-03
doc: Refactor coverage setup validation CY-6365

prcr:
Refactors the coverage documentation to group together all setup validation steps, ordered to make it easier for users to catch any setup issues as early as possible during the validation.

Together with https://github.com/codacy/codacy-coverage-reporter/pull/406, these improvements will give way to more detailed troubleshooting instructions for the most common coverage setup failure scenarios (see [CY-6263](https://codacy.atlassian.net/browse/CY-6263)).

### DIFF
--- a/docs/troubleshooting-common-issues.md
+++ b/docs/troubleshooting-common-issues.md
@@ -21,9 +21,9 @@ If the Codacy Coverage Reporter correctly uploaded your coverage report but the 
 
     ![Commits that must have coverage data](images/coverage-pr-commits.png)
 
-### Commit SHA hash detection {: id="commit-detection"}
+## Commit UUID detection {: id="commit-detection"}
 
-The Codacy Coverage Reporter automatically detects the commit SHA hash to associate with the coverage data from the following CI/CD platforms:
+The Codacy Coverage Reporter automatically detects the UUID or SHA hash of the current commit to associate with the coverage data when you're using one of the following CI/CD platforms:
 
 -   Appveyor
 -   AWS CodeBuild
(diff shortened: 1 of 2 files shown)


docs/troubleshooting-common-issues.md
```diff
@@ -48,7 +48,7 @@ The Codacy Coverage Reporter automatically detects the commit SHA hash to associ
 
 If the Codacy Coverage Reporter fails to detect the current commit from the CI workflow context, it will use the current commit from the local Git repository instead.
 
-However, you can also force using a specific commit SHA hash with the flag `--commit-uuid`. For example:
+However, you can also force using a specific commit UUID with the flag `--commit-uuid`. For example:
```
mrfyda, line 51: this change makes it consistent with the option, but SHA or hash seems to be the name use by git, while uuid is internal to Codacy 🤔

prcr: I was also checking that, and "commit SHA-1 hash" seems to be the most correct term. It's also [what GitHub uses](https://docs.github.com/en/search-github/searching-on-github/searching-commits#search-by-hash), for example.

Now that you mention this, I think it makes more sense to mention "commit SHA-1 hash" consistently even though the specific Codacy Coverage Reporter flag is `--commit-uuid`. This way we're using the same language that developers are already familiar with. :+1:

prcr: I pushed a commit to make the changes to the docs in this pull request and also opened https://github.com/codacy/codacy-coverage-reporter/pull/408 to suggest changing the way we refer to the SHA-1 hash in the Codacy Coverage Reporter outputs.

PR #408, opened 2022-08-03
docs: Change commit UUID to commit SHA-1 hash

prcr:
@mrfyda and I reached the conclusion that the Codacy Coverage Reporter is using the internal Codacy lingo "commit UUID" when referring to what's known by most developers as the "commit SHA-1 hash".

This is a proposal to change the way we refer to the SHA-1 hash in the CLI output of the Codacy Coverage Reporter.


pedrocodacy (approved):
LGTM

PR #418, opened 2022-10-04
feature: warn user when multiple languages are detected IO-52

pedrocodacy:
### Scope

For each coverage report file, the coverage reporter will:
- look through the files that are referenced in the report and extract its language by looking at file extensions
- if there are many languages, it will pick the first one and report all coverage as that language

This is problematic since it is not visible to the users (they just call our script) and it is possible that a report contains results for multiple languages.  

If you set the `-l` parameter it will override this behavior, but you must send the same report multiple times, one per language.

### This PR

When we make the decision to pick a specific language, we now output:
- that it is happening, 
- in what file, 
- what other languages were detected, and 
- a link to the relevant documentation.

### DIFF
--- a/src/main/scala/com/codacy/rules/ReportRules.scala
+++ b/src/main/scala/com/codacy/rules/ReportRules.scala
@@ -43,7 +43,7 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
               transform(coverageResult.report)(finalConfig)
           }
           _ <- storeReport(report, file)
-          language <- guessReportLanguage(finalConfig.languageOpt, report)
+          language <- guessReportLanguage(finalConfig.languageOpt, report, file.getAbsolutePath)
           success <- sendReport(report, language, finalConfig, commitUUID, file)
         } yield {
           success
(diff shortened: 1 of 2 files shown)


src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -194,18 +194,33 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
...
           case None => Left("Can't guess the report language")
-          case Some(value) => Right(value.name)
+          case Some(language) =>
+            if (reportLanguages.size > 1) {
+              logger.warn(s"""
+                   |Multiple languages detected in $reportFilePath (${reportLanguages.mkString(",")}).
+                   |  This run will report coverage for the $language language.
+                   |  To make sure you are uploading this report for all languages,
+                   |  check https://docs.codacy.com/coverage-reporter/uploading-coverage-in-advanced-scenarios/#multiple-languages
+                   |  for further details.""".stripMargin)
```
prcr, line 213: Suggestion to make the warning message shorter and reinforce that **only** the "elected" language is uploaded:

```suggestion
                   |Multiple languages detected in $reportFilePath (${reportLanguages.mkString(",")}).
                   |  This run will only upload coverage for the $language language.
                   |  To make sure that you upload coverage for all languages in the report,
                   |  see https://docs.codacy.com/coverage-reporter/uploading-coverage-in-advanced-scenarios/#multiple-languages""".stripMargin)
```

rubencodacy (dismissed):

prcr (approved):
:rocket:

PR #419, opened 2022-10-08
Update alternative run examples to be interactive

JBallin:
The [current ("Bash") examples](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#bash-script) don't show how to run the script interactively, so that users can configure the reporter using the CLI options.

I mainly want to document how to run the reporter on Alpine (without installing `bash`). I haven't tested that my update works on Ubuntu, but assuming this works it would be the same as the [regular run method](https://docs.codacy.com/coverage-reporter/#uploading-coverage)...which means we can possibly remove it entirely?

### DIFF
--- a/docs/alternative-ways-of-running-coverage-reporter.md
+++ b/docs/alternative-ways-of-running-coverage-reporter.md
@@ -19,9 +19,7 @@ The recommended way to run the Codacy Coverage Reporter is by using the [self-co
 -   On Alpine Linux, run:
 
     ```sh
-    wget -q https://coverage.codacy.com/get.sh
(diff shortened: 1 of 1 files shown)


nicklem:
Thank you for your suggestions.

We've [tested and applied them](https://github.com/codacy/docs/pull/1529), although you won't find them in this repo anymore—we recently moved the coverage reporter documentation to our main documentation repo.
For completeness, we decided to keep the command line for Ubuntu on the Alternatives page, even if that introduces some duplication.

I'm closing this pull request. You're very welcome to keep providing feedback on our docs! 🚀

PR #426, opened 2022-11-09
doc: Suggest rebasing branch to fix missing coverage variation

prcr:
If the common ancestor commit between a pull request and the target branch is missing coverage data, developers can rebase the branch to ensure that Codacy can calculate both coverage metrics.


rubencodacy (approved):
LGTM

PR #434, opened 2022-11-23
IO-142-Remove total field from coverage parser implementation 

OMaluhaCodacy:


coverage-parser/src/main/scala/com/codacy/parsers/implementation/JacocoParser.scala
```diff
@@ -1,72 +1,47 @@
...
 
   override val name: String = "Jacoco"
 
   private val ReportTag = "report"
 
   override def parse(projectRoot: File, reportFile: File): Either[String, CoverageReport] =
-    parseReport(reportFile, s"Could not find top level <$ReportTag> tag") {
-      parseReportNode(projectRoot, _)
+    parseReport(reportFile, s"Could not find top level <$ReportTag> tag") { node =>
+      Right(parseReportNode(projectRoot, node))
```
LazyBun, line 18: Nitpick: This could be: `parseReportNode(projectRoot, node).asRight`

You just would have to add `import cats.implicits._` if it's not there already

LazyBun: The `_` could be left in then (I think)

coverage-parser/src/main/scala/com/codacy/parsers/implementation/LCOVParser.scala
```diff
@@ -31,12 +31,12 @@ object LCOVParser extends CoverageParser {
   }
 
   private def parseLines(reportFile: File, lines: Iterator[String]): Either[String, CoverageReport] = {
-    val coverageFileReports =
+    val coverageFileReports: Either[String, Seq[CoverageFileReport]] =
```
LazyBun, line 34: Is it necessary to define the type there?

coverage-parser/src/test/scala/com/codacy/parsers/CloverParserTest.scala
```diff
@@ -25,7 +25,7 @@ class CloverParserTest extends WordSpec with Matchers with EitherValues {
       }
 
       "the report is not in the Clover format" in {
-        // Arrange
+        // Arranges
```
LazyBun, line 28: Unnecessary `s` I think

coverage-parser/src/test/scala/com/codacy/parsers/CloverParserTest.scala
```diff
@@ -156,7 +131,7 @@ class CloverParserTest extends WordSpec with Matchers with EitherValues {
     "return a report with the expected file coverage" in {
       // Arrange
       val filePath = "src/Codacy/Coverage/Parser/Parser.php"
-      val expectedFileCoverage = 33
+      val expectedFileCoverage = 0
```
LazyBun, line 134: This seems like a weird change to make in PR that just removes the `total`. Just want to make sure that it's intentional

LazyBun: Oh, it's that the total field is defaulted to 0. I see. Leaving this comment for posterity.

LazyBun (approved):
Code looks good, can't say whether anything else was supposed to be removed though. From my side - approved

coverage-parser/src/main/scala/com/codacy/parsers/implementation/LCOVParser.scala
OMaluhaCodacy: I don't think so, so deleted:)

coverage-parser/src/test/scala/com/codacy/parsers/CloverParserTest.scala
OMaluhaCodacy: removed

coverage-parser/src/test/scala/com/codacy/parsers/CloverParserTest.scala
OMaluhaCodacy: Test deleted as useless

coverage-parser/src/main/scala/com/codacy/parsers/implementation/JacocoParser.scala
LazyBun: @OMaluhaCodacy told me on Slack that this repo does not use cats. If that's the case then there is no point in adding this for something as unimportant as this

rubencodacy (approved):

PR #441, opened 2022-12-07
Change the get.sh to pin the coverage reporter version to a working o…

rubencodacy:
…ne for self hosted instances

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -162,9 +162,18 @@ download_reporter() {
 
 os_name=$(uname)
 
+# This version should be one that matches the latest self hosted release.
+SELF_HOSTED_CODACY_REPORTER_VERSION="13.10.15"
+
 # Find the latest version in case is not specified
 if [ -z "$CODACY_REPORTER_VERSION" ] || [ "$CODACY_REPORTER_VERSION" = "latest" ]; then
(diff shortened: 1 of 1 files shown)


get.sh
```diff
@@ -162,9 +162,18 @@ download_reporter() {
...
 os_name=$(uname)
 
+# This version should be one that matches the latest self hosted release.
+SELF_HOSTED_CODACY_REPORTER_VERSION="13.10.15"
+
 # Find the latest version in case is not specified
 if [ -z "$CODACY_REPORTER_VERSION" ] || [ "$CODACY_REPORTER_VERSION" = "latest" ]; then
-    CODACY_REPORTER_VERSION=$(download_stdout "https://artifacts.codacy.com/bin/codacy-coverage-reporter/latest")
+    # In case of a self hosted installation, pin a version to the latest released self hosted version working coverage reporter
+    if [ -n "$CODACY_API_BASE_URL" ] && [ "$CODACY_API_BASE_URL" != "https://api.codacy.com" ]; then
```
pedrocodacy, line 171: can we match on at least the `dev` and `staging` envs urls here? 

If we dont we wont be able to test new changes to the coverage reporter without setting the `CODACY_REPORTER_VERSION` when sending results to those envs

pedrocodacy: also please add `app.codacy.com` as a match as well since some users may have set it instead of `api` since it works the same

rubencodacy: Done

PR #447, opened 2023-02-10
TS-221 Parse raw go coverage report taking into account multiple files

rubencodacy:
… Add explanation about parser limitation/problem


stefanvacareanu7 (dismissed):

rubencodacy:
This does not have any fix, just an explanation about one limitation of the tool (go coverage)

PR #448, opened 2023-02-23
Exclude using internal classes directly TS-261

stefanvacareanu7:


stefanvacareanu7:
I notice another place where we use an internal class, [here](https://github.com/codacy/codacy-coverage-reporter/blob/6d0b6a7ee9d448e8461ee4cfcedfeadf28dd3b6e/build.sbt#L73), but this is a different case, I am not sure if this is affecting the current blocker.

rubencodacy (dismissed):

PR #484, opened 2023-12-15
wip: only files in git will be considered in the report, file name wi…

pedrobpereira:
…ll be according to git file structure ALA-773

### DIFF
--- a/coverage-parser/src/main/scala/com/codacy/parsers/implementation/JacocoParser.scala
+++ b/coverage-parser/src/main/scala/com/codacy/parsers/implementation/JacocoParser.scala
@@ -1,61 +1,46 @@
 package com.codacy.parsers.implementation
 
 import java.io.File
-
 import com.codacy.api._
-import com.codacy.parsers.util.TextUtils
+import com.codacy.parsers.util.FileNameMatcher
 import com.codacy.parsers.{CoverageParser, XmlReportParser}
+import wvlet.log.LogSupport
 
 import scala.xml.{Elem, Node, NodeSeq}
 
 private case class LineCoverage(missedInstructions: Int, coveredInstructions: Int)
 
-object JacocoParser extends CoverageParser with XmlReportParser {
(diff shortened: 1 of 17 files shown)


coverage-parser/src/main/scala/com/codacy/parsers/util/FileNameMatcher.scala
```diff
@@ -0,0 +1,36 @@
+package com.codacy.parsers.util
+
+import java.nio.file.Paths
+import scala.util.Try
+
+object FileNameMatcher {
+
+  def matchAndReturnName(filename: String, fileNames: Seq[String]): Option[String] = {
+
+    fileNames.find(name => isTheSameFile(filename, name))
```
pedrobpereira, line 10: To achieve parity with the already implemented mechanisms in our code base (coverage-service, engine) there is still missing a part where we sort by **similarity** (see code base)

coverage-parser/src/main/scala/com/codacy/parsers/implementation/JacocoParser.scala
```diff
@@ -1,61 +1,46 @@
...
-      .getOrElse {
-        Left("Could not retrieve total percentage of coverage.")
-      }
+  private def parseReportNode(report: NodeSeq, acceptedFiles: Seq[String]): Either[String, CoverageReport] = {
+    val filesCoverage = for {
+      pkg <- report \\ "package"
+      packageName = pkg \@ "name"
+      sourceFile <- pkg \\ "sourcefile"
+      filename = s"$packageName/${sourceFile \@ "name"}"
+      actualName <- FileNameMatcher
```
pedrobpereira, line 34: This should be moved to a more generic place, after the reports are done, so that we don't have to implement for each parser

pedrobpereira:
Will be tackled here https://github.com/codacy/codacy-coverage-reporter/pull/485

PR #485, opened 2023-12-20
feature: only files in git will be considered in the report ALA-792

pedrobpereira:
file name will be according to git file structure ALA-792

### DIFF
--- a/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
+++ b/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
@@ -2,9 +2,10 @@ package com.codacy.api.helpers.vcs
 
 import java.io.File
 import java.util.Date
-
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.lib.{Repository, RepositoryBuilder}
+import org.eclipse.jgit.revwalk.RevWalk
+import org.eclipse.jgit.treewalk.TreeWalk
 
 import scala.collection.JavaConverters._
 import scala.util.Try
@@ -35,6 +36,29 @@ class GitClient(workDirectory: File) {
       val authorIdent = headRev.getAuthorIdent
 
       CommitInfo(headRev.getName, authorIdent.getName, authorIdent.getEmailAddress, authorIdent.getWhen)
+    }
+  }
+
+  def getRepositoryFileNames(commitSha: String): Try[Seq[String]] = {
+    repositoryTry.map { rep =>
+      val git = new Git(rep)
+      val repo = git.getRepository
+      val commitId = repo.resolve(commitSha)
+      val revWalk = new RevWalk(repo)
+      val commit = revWalk.parseCommit(commitId)
+      val tree = commit.getTree
+      val treeWalk = new TreeWalk(repo)
+      treeWalk.addTree(tree)
+      treeWalk.setRecursive(true)
+
+      val result: Seq[String] =
+        if (treeWalk.next) {
+          Stream
+            .continually(treeWalk.getPathString)
+            .takeWhile(_ => treeWalk.next)
+        } else Seq.empty
+
+      result
     }
   }
 
--- a/src/main/scala/com/codacy/rules/ReportRules.scala
+++ b/src/main/scala/com/codacy/rules/ReportRules.scala
@@ -8,14 +8,15 @@ import com.codacy.model.configuration._
 import com.codacy.parsers.CoverageParser
 import com.codacy.plugins.api.languages.Languages
 import com.codacy.rules.commituuid.CommitUUIDProvider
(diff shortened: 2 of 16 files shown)


api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
```diff
@@ -38,4 +39,27 @@ class GitClient(workDirectory: File) {
...
+      val commitId = repo.resolve(commitSha)
+      val revWalk = new RevWalk(repo)
+      val commit = revWalk.parseCommit(commitId)
+      val tree = commit.getTree
+      val treeWalk = new TreeWalk(repo)
+      treeWalk.addTree(tree)
+      treeWalk.setRecursive(true)
+
+      val result: Seq[String] =
+        if (treeWalk.next) {
```
machadoit, line 55: interesting if, is it really needed? Don't the continually take care of it? Also, should this return Stream explicitly? To know that the caller needs to consume this?

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -318,8 +323,10 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
     }
   }
 
-  private def transform(report: CoverageReport)(config: ReportConfig): CoverageReport = {
-    val transformations = Set(new PathPrefixer(config.prefix))
+  private def transform(
+      report: CoverageReport
+  )(config: ReportConfig, acceptableFileNames: Seq[String]): CoverageReport = {
+    val transformations = Set(new PathPrefixer(config.prefix), new GitFileNameUpdaterAndFilter(acceptableFileNames))
```
machadoit, line 329: Be careful with the the Set here, probably we should change to a Seq. Since Set does not guarantee order, and we should guarantee that the GitFileNameUpdaterAndFilteris the last transformation

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -32,15 +33,19 @@ class ReportRules(coverageServices: => CoverageServices) extends LogSupport {
...
     val finalConfig = config.copy(partial = partial)
+
+    val gitFiles = gitFileFetcher.forCommit(commitUUID)
+
     files
       .map { file =>
         logger.info(s"Parsing coverage data from: ${file.getAbsolutePath} ...")
         for {
           _ <- validateFileAccess(file)
+          files <- gitFiles
```
machadoit, line 44: We can consider this for to allow for optional files maybe? So we fallback to not doing this new mechanism if anything fails

src/main/scala/com/codacy/transformation/GitFileNameUpdaterAndFilter.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.codacy.api.CoverageReport
+import wvlet.log.LogSupport
+
+class GitFileNameUpdaterAndFilter(acceptableFileNames: Seq[String]) extends Transformation with LogSupport {
+  override def execute(report: CoverageReport): CoverageReport = {
+    val fileReports = for {
+      fileReport <- report.fileReports
+      fileName <- FileNameMatcher
+        .matchAndReturnName(fileReport.filename, acceptableFileNames)
+        .map(Some(_))
```
machadoit, line 11: Strange code here, maybe an .orElse on the log, is that what you want?

api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
pedrobpereira: Yeah, I think it is needed, I'll test this again without, but I added it because it was failing without the first next. Since return type is typified, I don't think we have to worry about that right, whoever is consuming will only interact with a Seq.

src/main/scala/com/codacy/transformation/GitFileNameUpdaterAndFilter.scala
manufacturist: I think it looks bloated even with .orElse. Maybe move it outside:

```scala
  [...]
    fileName <- matchFileName(fileReport)
  [...]

  private def matchFileName(fileReport: CoverageFileReport): Option[String] = {
    val maybeFileName = FileNameMatcher.matchAndReturnName(fileReport.filename, acceptableFileNames)

    if (maybeFileName.isEmpty)
      logger.warn(s"File: ${fileReport.filename} will be discarded and will not be considered for coverage calculation")

    maybeFileName
  }
```

machadoit (dismissed):
LGTM!

pedrobpereira:
Converted to Draft again to prevent someone merging this, we will be waiting till the 1st week of January to go ahead with this.

machadoit (approved):

PR #486, opened 2024-01-10
feature: stops failure when at least one report is sent ALA-799

pedrobpereira:
stops returning failure when at least one report is sent ALA-799


machadoit (approved):
LGTM!

PR #487, opened 2024-01-23
doc: Update error messages DOCS-645

nicklem:


claudiacarpinteiro (approved):
LGTM! 👍

src/main/scala/com/codacy/rules/ReportRules.scala
```diff
@@ -120,7 +120,7 @@ class ReportRules(coverageServices: => CoverageServices, gitFileFetcher: GitFile
     */
   private[rules] def storeReport(report: CoverageReport, file: File) = {
     if (report.fileReports.isEmpty)
-      Left(s"The provided coverage report ${file.getAbsolutePath} generated an empty result.")
+      Left(s"The uploaded coverage report ${file.getAbsolutePath} is empty or contains no data for repository files.")
```
DMarinhoCodacy, line 123: The coverage report was not uploaded in this part yet. So I suggest to change to "The parsed report" or "The generated report". Do you agree @pedrobpereira ?

DMarinhoCodacy (changes requested):

src/main/scala/com/codacy/rules/ReportRules.scala
pedrobpereira: Yeah better to name it something else, because uploaded seems to be pointing to the report provided by the user, and that one might not be empty, only the one we generate is empty due to file matching. I would suggest "generated report" as @DMarinhoCodacy pointed out.

nicklem: Generated sounds good to me too. Let's go with that one

```suggestion
      Left(s"The generated coverage report ${file.getAbsolutePath} is empty or contains no data for repository files.")
```

DMarinhoCodacy (approved):
LGTM!

PR #505, opened 2024-05-31
use toLowerCase to compare all files and paths

DMarinhoCodacy:


src/main/scala/com/codacy/transformation/FileNameMatcher.scala
```diff
@@ -20,13 +20,13 @@ object FileNameMatcher {
     getFilenameFromPath(file).equalsIgnoreCase(getFilenameFromPath(covFile))
 
   private def haveSamePath(file: String, covFile: String): Boolean =
-    file.equalsIgnoreCase(covFile)
+    file.toLowerCase.equalsIgnoreCase(covFile.toLowerCase)
```
machadoit, line 23: The method equalsIgnore case should already take this in consideration no?

DMarinhoCodacy: Yes, you don't need to use toLowerCase for that.

src/main/scala/com/codacy/transformation/FileNameMatcher.scala
```diff
@@ -7,20 +7,20 @@ object FileNameMatcher {
...
     fileNames
-      .filter(name => isTheSameFile(filename, name))
+      .filter(name => isTheSameFile(filename.toLowerCase, name.toLowerCase))
       .sortBy(name => Math.abs(filename.length - name.length))
       .headOption
   }
 
   def getFilenameFromPath(filename: String): String = {
-    Try(Paths.get(filename).getFileName.toString).getOrElse(filename)
+    Try(Paths.get(filename).getFileName.toString.toLowerCase).getOrElse(filename.toLowerCase)
```
machadoit, line 16: Do you know if this Paths.get(filename) should rely on the lower case or if they return the file on either case?
Is there a possibility to add a test for this?

Anyway I am fine moving forward if this unblocks a customer, but adding a follow up PR with a test could be cool

machadoit (dismissed):

machadoit:
I guess I would expect some assertion on the specific file, but I also didn't touch this repo much, and see some similar tests, so LGTM!

machadoit (approved):

PR #507, opened 2024-06-14
breaking: Update to Scala `3.3.3`

lolgab:
Scala 3 case-app doesn't depend on shapeless anymore, which should reduce the binary size
This also updates dependencies to their Scala 3 compatible counterpart
- `case-app` is updated to `2.1.0-M28`
- `airframe-log` is updated to `24.6.0`
- `mockito` is replaced with `scalamock` which supports scala 3
- `scalatest` is updated to `3.2.18`
- `play-json` is updated to `2.10.5`
- `codacy-plugins-api` is updated to `8.1.4`
- `scala-xml` is updated to `2.3.0`

### DIFF
--- a/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
+++ b/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
@@ -7,7 +7,7 @@ import org.eclipse.jgit.lib.{Repository, RepositoryBuilder}
 import org.eclipse.jgit.revwalk.RevWalk
 import org.eclipse.jgit.treewalk.TreeWalk
 
-import scala.collection.JavaConverters._
+import scala.jdk.CollectionConverters._
 import scala.util.Try
 
 case class CommitInfo(uuid: String, authorName: String, authorEmail: String, date: Date)
@@ -53,9 +53,10 @@ class GitClient(workDirectory: File) {
 
       val result: Seq[String] =
         if (treeWalk.next) {
-          Stream
+          Iterator
             .continually(treeWalk.getPathString)
             .takeWhile(_ => treeWalk.next)
+            .toSeq
         } else Seq.empty
 
       result
--- a/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
+++ b/api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
@@ -53,10 +53,9 @@ class GitClient(workDirectory: File) {
 
       val result: Seq[String] =
         if (treeWalk.next) {
-          Iterator
+          LazyList
             .continually(treeWalk.getPathString)
             .takeWhile(_ => treeWalk.next)
-            .toSeq
         } else Seq.empty
 
       result
--- a/build.sbt
+++ b/build.sbt
@@ -1,23 +1,21 @@
-inThisBuild(
-  Seq(
-    scalaVersion := "2.13.13",
-    scalacOptions := Seq("-deprecation", "-feature", "-unchecked", "-Xlint", "-Xfatal-warnings")
-  )
-)
+inThisBuild(Seq(scalaVersion := "3.3.3"))
+
+def commonSettings =
+  Seq(scalacOptions := {
+    val toFilter = Set("-deprecation:false")
+    scalacOptions.value.filterNot(toFilter) ++ Seq("-deprecation")
+  })
 
 name := "codacy-coverage-reporter"
 
 // Runtime dependencies
 libraryDependencies ++= Seq(
-  "com.github.alexarchambault" %% "case-app" % "2.1.0-M26",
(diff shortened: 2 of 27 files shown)


api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
```diff
@@ -53,9 +53,10 @@ class GitClient(workDirectory: File) {
 
       val result: Seq[String] =
         if (treeWalk.next) {
-          Stream
+          Iterator
```
machadoit, line 56: Before this was lazy, but now with the `.toSeq` it is not. Maybe we want to use `LazyList` here to maintain the semantics?

build.sbt
```diff
@@ -1,23 +1,21 @@
...
+    scalacOptions.value.filterNot(toFilter) ++ Seq("-deprecation")
+  })
 
 name := "codacy-coverage-reporter"
 
 // Runtime dependencies
 libraryDependencies ++= Seq(
-  "com.github.alexarchambault" %% "case-app" % "2.1.0-M26",
-  "org.wvlet.airframe" %% "airframe-log" % "22.3.0"
+  "com.github.alexarchambault" %% "case-app" % "2.1.0-M28",
```
machadoit, line 13: 🙃

build.sbt
```diff
@@ -72,24 +70,30 @@ nativeImageOptions := Seq(
...
+
 lazy val apiScala = project
   .in(file("api-scala"))
   .settings(
+    commonSettings,
     libraryDependencies ++= Seq(
-      "com.typesafe.play" %% "play-json" % "2.8.2",
-      "org.scalaj" %% "scalaj-http" % "2.4.2",
+      "com.typesafe.play" %% "play-json" % "2.10.5",
+      ("org.scalaj" %% "scalaj-http" % "2.4.2").cross(CrossVersion.for3Use2_13),
```
machadoit, line 83: What is the ideia here? That lib is not available for Scala3, and with this we are relying on the one from 2.13? (It also seems deprecated, so I guess we will want to swap at some point)

src/test/scala/com/codacy/rules/ReportRulesSpec.scala
```diff
@@ -67,22 +69,22 @@ class ReportRulesSpec extends WordSpec with Matchers with PrivateMethodTester wi
...
       }
 
       "it is not able to parse report file" in {
-        val coverageServices = mock[CoverageServices]
-        val gitFileFetcher = mock[GitFileFetcher]
+        val coverageServices = stub[CoverageServices]
+        val gitFileFetcher = stub[GitFileFetcher]
 
-        gitFileFetcher.forCommit(any[String]).shouldReturn(Right(Seq.empty))
+        (gitFileFetcher.forCommit _).when(*).returns(Right(Seq.empty))
```
machadoit, line 87: Such a strange syntax imo, but it is what it is

build.sbt
```diff
@@ -1,23 +1,21 @@
...
-    scalaVersion := "2.13.13",
-    scalacOptions := Seq("-deprecation", "-feature", "-unchecked", "-Xlint", "-Xfatal-warnings")
-  )
-)
+inThisBuild(Seq(scalaVersion := "3.3.3"))
+
+def commonSettings =
+  Seq(scalacOptions := {
+    val toFilter = Set("-deprecation:false")
+    scalacOptions.value.filterNot(toFilter) ++ Seq("-deprecation")
```
machadoit, line 6: What is happening here and why do we need it?

build.sbt
lolgab: Exactly!
I tried to swap with sttp once and I created an outage, but I will give another try with a newer and better maintained library. Fortunately Sbt can check for incompatibilities with the mix of `_2.13` and `_3` dependencies, so we are totally safe here.

build.sbt
lolgab: There were two issues that were failing compilation because of `-Xfatal-warnings` and couldn't be seen because of the `-deprecation:false` setting in `codacy-sbt`. Maybe we need to revisit `scalacOptions` in codacy-sbt but until that day this removes that option and adds `-deprecation`.

build.sbt
lolgab: 🤷‍♂️

src/test/scala/com/codacy/rules/ReportRulesSpec.scala
lolgab: Yes, Scalamock stretches Scala's syntax to be able to mock at compile time and not using reflection, so I'm happy enough.

api-scala/src/main/scala/com/codacy/api/helpers/vcs/GitClient.scala
lolgab: Good point! Fixing it.

machadoit (approved):
LGTM!

PR #509, opened 2024-07-18
Make the count property optional on line node for clover parser

hjrocha:


.circleci/config.yml
```diff
@@ -248,13 +248,16 @@ workflows:
             export PATH=$HOME/.musl/x86_64-linux-musl-native/bin:$PATH
             sbt "assembly;nativeImage"
             mkdir -p ~/workdir/tmp-artifacts
+            ls target
```
machadoit, line 251: Are those left here on purpose?

coverage-parser/src/main/scala/com/codacy/parsers/implementation/CloverParser.scala
```diff
@@ -71,7 +71,8 @@ object CloverParser extends CoverageParser with XmlReportParser {
     fileLineTags.foldLeft[Either[String, Map[Int, Int]]](Right(Map.empty[Int, Int])) {
       case (left: Left[_, _], _) => left
 
-      case (Right(lines), line) if (line \@ "type") == "stmt" || (line \@ "type") == "cond" =>
+      case (Right(lines), line)
+          if ((line \@ "type") == "stmt" || (line \@ "type") == "cond") && (line \@ "count").nonEmpty =>
```
machadoit, line 75: So you only want to consider lines that have count defined? Seems a bit weird that there is an or to check for cond, but then you added that check as well 🤔 

Seems that if the intention is not to count me, that we should revert the https://github.com/codacy/codacy-coverage-reporter/pull/420/files which seems that was added to count those lines?

machadoit (approved):
If you are confident on the change, feel free to move forward, but would check the mentioned PR and if we want to make this change.

Also, is there any ticket you can link to the PR for future reference on why we want to make the change?

manufacturist (approved):

PR #516, opened 2025-01-13
[DO NOT MERGE YET] Add Skip SSL Flag

heliocodacy:


api-scala/src/main/scala/com/codacy/api/client/CodacyClient.scala
```diff
@@ -44,8 +47,10 @@ class CodacyClient(
 
       val httpRequest = timeoutOpt match {
         case Some(timeout) =>
-          Http(url).timeout(connTimeoutMs = timeout.connTimeoutMs, readTimeoutMs = timeout.readTimeoutMs)
-        case None => Http(url)
+          Http(url)
+            .timeout(connTimeoutMs = timeout.connTimeoutMs, readTimeoutMs = timeout.readTimeoutMs)
+            .options(httpOptions)
```
machadoit, line 52: Assuming that the httpOptions when empty does not override anything previously set, works for me! Just validate that by default there are no options. Meaning, confirming that `.options(Seq.emtpy)` is the same as not doing anything

machadoit (approved):

